PR #12, opened 2018-08-24
Allow azure_roles to be specified as a string or JSON

kalafut:
I decided to just make this work with the existing TypeSlice for now. We can also look at adding a better type to Vault (or making the existing ones handle strings too, for example), but avoiding changes to Vault _right now_ seemed preferable.

### DIFF
--- a/path_roles.go
+++ b/path_roles.go
@@ -13,6 +13,7 @@ import (
 	"github.com/hashicorp/vault/helper/jsonutil"
 	"github.com/hashicorp/vault/logical"
 	"github.com/hashicorp/vault/logical/framework"
+	"github.com/mitchellh/mapstructure"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


jefferai:
I'm 👎 on this. I'd rather see a TypeJSON that can more intelligently accept various inputs, similarly to how we accept multiple types of possible inputs for database SQL.

That's separate even from the question of whether this is even really an issue. The CLI will stringify JSON for you if you're sourcing it from a file, and from any API it's pretty trivial to marshal JSON. It's pretty much only an issue if you're using curl, and even then there are things like jq that can help.

kalafut:
Following some more discussion on this, there's a proposal to have TypeMap and TypeSlice handle string input: https://github.com/hashicorp/vault/pull/5182

I'll update this PR based on what lands in Vault.

kalafut:
RFC under discussion: https://docs.google.com/document/d/1DYj1rKENgDsrU_SedJM8GNoYC53WRqORhNiOAnbrwCo

kalafut:
Retiring this until/if TypeJSON or some variant lands in Vault.

PR #14, opened 2018-10-01
Add new "static service principal" credential type

kalafut:
Still TODO
- determine passwords per app limit
- docs

### DIFF
--- a/backend_test.go
+++ b/backend_test.go
@@ -69,6 +69,7 @@ func getTestBackend(t *testing.T, initConfig bool) (*azureSecretBackend, logical
 type mockProvider struct {
 	subscriptionID            string
 	applications              map[string]bool
+	passwords                 map[string]bool
 	failNextCreateApplication bool
 }
 
@@ -76,6 +77,7 @@ func newMockProvider() AzureProvider {
 	return &mockProvider{
 		subscriptionID: generateUUID(),
 		applications:   make(map[string]bool),
+		passwords:      make(map[string]bool),
 	}
 }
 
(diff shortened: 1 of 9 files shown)


kalafut:
This PR will be updated to remove the `credential_type` parameter from roles (which was just added in the PR). The presence of `application_object_id` will be the indicator that a new application & SP should not be created.

backend_test.go
```diff
@@ -148,15 +150,45 @@ func (m *mockProvider) CreateApplication(ctx context.Context, parameters graphrb
...
+	for _, v := range *parameters.Value {
+		m.passwords[*v.KeyID] = true
+	}
+
+	return autorest.Response{}, nil
+}
+
+func (m *mockProvider) ListApplicationPasswordCredentials(ctx context.Context, applicationObjectID string) (result graphrbac.PasswordCredentialListResult, err error) {
+	var creds []graphrbac.PasswordCredential
+	for keyID, _ := range m.passwords {
```
tyrannosaurus-becks: FWIW, when you're ranging over a map and you only want the key but don't care about the value, you can just do:
```
for key := range someMap {
  // do stuff
}
```

client.go
```diff
@@ -143,6 +143,92 @@ func (c *client) createSP(
...
+	// Key IDs are not secret, and they're a convenient way for an operator to identify Vault-generated
+	// passwords. These must be UUIDs, so the three leading bytes will be used as an indicator.
+	keyID = "ffffff" + keyID[6:]
+
+	password, err := uuid.GenerateUUID()
+	if err != nil {
+		return "", "", err
+	}
+
+	now := time.Now()
```
tyrannosaurus-becks: Pretty sure this will grab the current time in whatever timezone the server is in that's hosting the application. If the server were in US Eastern time, would everything still work as expected? Especially if the remote Azure server were set to a different timezone?

backend_test.go
kalafut: Oh yeah, thanks.

client.go
kalafut: It will probably work OK since what gets sent is a full RFC3339 timestamp, and one of their samples shows time.Now() being used. But the `az` CLI tool sends UTC, and passwords created in the portal get UTC, so UTC it is. 🙂

tyrannosaurus-becks: Cool cool, I figured it did something like that but just wanted to ask. :-)

client.go
```diff
@@ -143,6 +143,91 @@ func (c *client) createSP(
...
+// addAppPassword adds a new password to an App's credentials list.
+func (c *client) addAppPassword(ctx context.Context, appObjID string, duration time.Duration) (string, string, error) {
+	keyID, err := uuid.GenerateUUID()
+	if err != nil {
+		return "", "", err
+	}
+
+	// Key IDs are not secret, and they're a convenient way for an operator to identify Vault-generated
+	// passwords. These must be UUIDs, so the three leading bytes will be used as an indicator.
+	keyID = "ffffff" + keyID[6:]
```
chrishoffman: why not `vault1` or something like that?

client.go
```diff
@@ -143,6 +143,91 @@ func (c *client) createSP(
...
+
+	// Add and save credentials
+	curCreds = append(curCreds, cred)
+
+	if _, err := c.provider.UpdateApplicationPasswordCredentials(ctx, appObjID,
+		graphrbac.PasswordCredentialsUpdateParameters{
+			Value: &curCreds,
+		},
+	); err != nil {
+		if strings.Contains(err.Error(), "size of the object has exceeded its limit") {
```
chrishoffman: This is nice to handle this, even if it eventually breaks due to response changes.

client.go
kalafut: keyID must be a UUID string. `ffffff` was chosen since they'll sort to the bottom of the list in the Portal view. I'm assuming there might be lots of Vault keys but just a few manually created ones, which should be more prominent.

PR #16, opened 2018-10-24
Fix duplicate role definition check

### LINKED ISSUE #15: duplicate role_id
Hello, 
I want to configure a  role with the same azure role on different scope. 

```
vault write providers/azure/roles/my-role ttl=30m azure_roles=-<<EOF
[
    {
        "role_name": "Contributor",
        "scope": "/subscriptions/XXXXXXXX/resourceGroups/RG1"                                                                                                 
    },
    {
        "role_name": "Contributor",
        "scope": "/subscriptions/XXXXXXXX/resourceGroups/RG2"
    }
]
EOF

Code: 400. Errors:

* duplicate role_id:'/subscriptions/XXXXXXXX/providers/Microsoft.Authorization/roleDefinitions/b24988ac-6180-42a0-ab88-20f7382dd24c'
```

How I can do this ? 

Thanks

kalafut:
(role_id, scope) must be unique, not role_id alone.

Fixes #15

### DIFF
--- a/path_roles.go
+++ b/path_roles.go
@@ -166,7 +166,7 @@ func (b *azureSecretBackend) pathRoleUpdate(ctx context.Context, req *logical.Re
 		role.AzureRoles = parsedRoles
 	}
 
-	roleSet := make(map[azureRole]bool)
+	roleSet := make(map[string]bool)
 	for _, r := range role.AzureRoles {
 		var roleDef authorization.RoleDefinition
 		if r.RoleID != "" {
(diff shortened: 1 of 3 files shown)


path_roles.go
```diff
@@ -166,7 +166,7 @@ func (b *azureSecretBackend) pathRoleUpdate(ctx context.Context, req *logical.Re
 		role.AzureRoles = parsedRoles
 	}
 
-	roleIDs := make(map[string]bool)
+	roleSet := make(map[azureRole]bool)
```
tyrannosaurus-becks:
```
type azureRole struct {
	RoleName string `json:"role_name"` // e.g. Owner
	RoleID   string `json:"role_id"`   // e.g. /subscriptions/e0a207b2-.../providers/Microsoft.Authorization/roleDefinitions/de139f84-...
	Scope    string `json:"scope"`     // e.g. /subscriptions/e0a207b2-...
}
```
I notice that the PR says it wants RoleID and Scope to be unique, but actually, if I'm reading this right, _any of_ these three fields could be different and it would result in an azureRole not being equal to any already in the map. Is that what you want? Also, any fields added in the future could change behavior around equality comparisons. Would it make more sense to create a custom equality comparison to lock down the behaviour?

kalafut: I probably could have been clearer in the PR. I did mean the combo of RoleID+Scope needs to be unique, so the current behavior works (RoleName is effectively joined to RoleID and isn't a unique variable). But your comment about expansion makes sense. I think I'll use a key derived from the struct so changes don't message things up.

PR #24, opened 2019-04-18
Revert Azure SDK to pre-go-modification version to fix SP create errors

kalafut:
The update to go modules brought in much newer Azure SDK and Autorest versions, along with a one-line code change to allow building. The resulting plugin doesn't successfully create new Service Principals (the integ tests fail), and the full scope of SDK changes from v19->27 needs to be reviewed.  This PR restores the two libraries to their pre-go-mod versions.


kalafut:
Azure SDK v27-28 have flaws. Opening a separate PR to go to v29.

PR #31, opened 2019-09-01
Add Dynamic Service Principal to AAD Groups

dadgar:
This PR supports adding the dynamic service principal to an Azure Active Directory group. The group can be specified by name or by object_id. The required permissions to add to a group are `Azure Active Directory Graph / Directory.ReadWrite.All`. I have tested this by building the plugin and generating a service principal, seeing it be added to the group, and removed when I revoke the lease.

For the error codes, I tested by requesting non-existing service providers and mapping the errors. I have not added anything to the integration test since I do not have access to that environment. 

Additional Role Config is as follows:
```
"azure_groups": [
 	{
 		"group_name": "", 
 		"object_id": "239b11fe-6adf-409a-b231-08b918e9de23"
 	},
 	{
 		"group_name": "bar",
 		"object_id": "" // Looked up
 	}
]
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -353,7 +353,7 @@ func (c *client) removeGroupMemberships(ctx context.Context, servicePrincipalObj
 	return merr.ErrorOrNil()
 }
 
-// groupObjectIDs is a helper for creating converting a list of AzureGroup
+// groupObjectIDs is a helper for converting a list of AzureGroup
 // objects to a list of their object IDs.
 func groupObjectIDs(groups []*AzureGroup) []string {
 	groupIDs := make([]string, 0, len(groups))
--- a/client.go
+++ b/client.go
@@ -321,9 +321,8 @@ func (c *client) addGroupMemberships(ctx context.Context, sp *graphrbac.ServiceP
 					),
 				})
 
-			// TODO is this right?
 			// Propagation delays within Azure can cause this error occasionally, so don't quit on it.
-			if err != nil && strings.Contains(err.Error(), "PrincipalNotFound") {
+			if err != nil && strings.Contains(err.Error(), "Request_ResourceNotFound") {
 				return nil, false, nil
 			}
 
--- a/client.go
+++ b/client.go
@@ -306,9 +306,64 @@ func (c *client) unassignRoles(ctx context.Context, roleIDs []string) error {
 	return merr.ErrorOrNil()
 }
 
+// addGroupMemberships adds the service principal to the Azure groups.
+func (c *client) addGroupMemberships(ctx context.Context, sp *graphrbac.ServicePrincipal, groups []*AzureGroup) ([]string, error) {
+	groupsIDs := make([]string, 0, len(groups))
+	for _, group := range groups {
+		_, err := retry(ctx, func() (interface{}, bool, error) {
+			_, err := c.provider.AddGroupMember(ctx, group.ObjectID,
+				graphrbac.GroupAddMemberParameters{
+					URL: to.StringPtr(
+						fmt.Sprintf("https://graph.windows.net/%s/directoryObjects/%s",
+							c.settings.TenantID,
+							*sp.ObjectID,
+						),
+					),
+				})
+
+			// TODO is this right?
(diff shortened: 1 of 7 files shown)


client.go
```diff
@@ -306,11 +306,65 @@ func (c *client) unassignRoles(ctx context.Context, roleIDs []string) error {
 	return merr.ErrorOrNil()
 }
 
+// addGroupMemberships adds the service principal to the Azure groups.
+func (c *client) addGroupMemberships(ctx context.Context, sp *graphrbac.ServicePrincipal, groups []*AzureGroup) ([]string, error) {
+	groupsIDs := make([]string, 0, len(groups))
```
kalafut: Is returning this just a convenience, or is there some cases in which the information is different than what the caller already has?

client.go
```diff
@@ -306,11 +306,65 @@ func (c *client) unassignRoles(ctx context.Context, roleIDs []string) error {
...
+	}
+
+	return groupsIDs, nil
+}
+
+// removeGroupMemberships removes the passed service principal from the passed
+// groups. This is a clean-up operation that isn't essential to revocation. As
+// such, an attempt is made to remove all memberships, and not return
+// immediately if there is an error.
+func (c *client) removeGroupMemberships(ctx context.Context, servicePrincipalID string, groupIDs []string) error {
```
kalafut: Why pass in a string instead of `*graphrbac.ServicePrincipal` here? (re: consistency)

path_roles.go
```diff
@@ -40,6 +42,16 @@ type AzureRole struct {
 	Scope    string `json:"scope"`     // e.g. /subscriptions/e0a207b2-...
 }
 
+// AzureGroup is an Azure Active Directory Group
+// (https://docs.microsoft.com/en-us/azure/role-based-access-control/overview).
+// GroupName and ObjectID are both traits of the group. ObjectID is the unique
+// identifier, but GroupName is more useful to a human (thought it is not
```
kalafut:
```suggestion
// identifier, but GroupName is more useful to a human (though it is not
```

path_roles.go
```diff
@@ -166,6 +187,18 @@ func (b *azureSecretBackend) pathRoleUpdate(ctx context.Context, req *logical.Re
 		role.AzureRoles = parsedRoles
 	}
 
+	// Parse the Azure groups
+	if groups, ok := d.GetOk("azure_groups"); ok {
+		parsedGroups := make([]*AzureGroup, 0) // non-nil to avoid a "missing groups" error later
+
+		err := jsonutil.DecodeJSON([]byte(groups.(string)), &parsedGroups)
+		if err != nil {
+			return logical.ErrorResponse(fmt.Sprintf("error parsing Azure groups '%s': %s", groups.(string), err.Error())), nil
```
kalafut:
```suggestion
			return logical.ErrorResponse("error parsing Azure groups '%s': %s", groups.(string), err.Error()), nil
```

Applies elsewhere too. Syntactical sugar now exists for this common pattern.

client.go
kalafut: Oh, I see this is just from internal data strings, so that's probably fine. Though getting the word "Object" into the parameter name would be good, since ApplicationID vs. ObjectID is always error prone.

client.go
dadgar: Yeah it was just for convenience. Pulled it out into a dedicated helper

client.go
```diff
@@ -306,11 +306,75 @@ func (c *client) unassignRoles(ctx context.Context, roleIDs []string) error {
...
+	for _, id := range groupIDs {
+		if _, err := c.provider.RemoveGroupMember(ctx, servicePrincipalObjectID, id); err != nil {
+			merr = multierror.Append(merr, errwrap.Wrapf("error removing group membership: {{err}}", err))
+		}
+	}
+
+	return merr.ErrorOrNil()
+}
+
+// groupObjectIDs is a helper for creating converting a list of AzureGroup
```
kalafut:
```suggestion
// groupObjectIDs is a helper for converting a list of AzureGroup
```

PR #33, opened 2019-11-19
Update Azure dep to prevent conflicts with other updated deps

jefferai:


jefferai:
Note: please merge, not squash, when merging.

PR #34, opened 2019-11-19
Fix integration test

kalafut:
The core issue is that normally secret data is persisted via JSON, which ends up (as expected) removing type data (e.g. a string pointer becomes a JSON string and is still a string in the `map[string]interface{}` after being read. Test paths aren't going through that persistence flow which was causing some errors.

### DIFF
--- a/path_service_principal_test.go
+++ b/path_service_principal_test.go
@@ -11,7 +11,6 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 	log "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-uuid"
-	"github.com/hashicorp/vault/sdk/helper/jsonutil"
 	"github.com/hashicorp/vault/sdk/helper/logging"
 	"github.com/hashicorp/vault/sdk/logical"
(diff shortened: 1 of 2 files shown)


path_service_principal_test.go
```diff
@@ -700,3 +674,17 @@ func TestCredentialInteg(t *testing.T) {
...
 	})
 }
+
+// fakeSaveLoad will simulate the JSON encoding/decoding process that secrets will normally go
+// through. If not done, some of the secret data will retain richer data types (e.g. pointers)
+// than would normally be seen.
+func fakeSaveLoad(s *logical.Secret) {
+	enc, err := jsonutil.EncodeJSON(s)
+	if err != nil {
+		panic(err)
```
jasonodonnell: Is `panic` necessary here?  Should we just `t.error()` instead?

kalafut: Since this is well into "should never happen" (within the context of a test) territory, I'm ok with a loud failure and test halt. That said, I just discovered I have a helper file with a number of these so I moved it over there.

Issue #50, opened 2020-10-23
Support for Oauth Access Tokens


mdgreenfield:
Requesting a new `token` endpoint for retrieving access tokens.

Similar to `vault-plugin-secrets-gcp`, requests to create a Vault role would create the Azure App/ServicePrincipal, store the corresponding client credentials in the plugin's internal storage, and use those same client credentials to generate an oauth access token when a user calls `azure/token/:role`

fairclothjm:
Thank you for submitting this request! For others who are interested in this, please stick a 👍 on this issue. We’re currently developing an internal process to review and prioritize feature requests.

mdgreenfield:
Thanks @fairclothjm, I just opened a PR that implements this (I'm not sure why I didn't do that a while back). The PR will certainly need to be rebased and documentation added at a minimum. I'll be away from the computer next week but potentially could look into some of these things the following week.

fairclothjm:
@mdgreenfield That's great, thanks! Feel free to ping me on this issue when you get it updated!

mdgreenfield:
Hi @fairclothjm, I could use some feedback/thoughts on https://github.com/hashicorp/vault-plugin-secrets-azure/pull/147#issuecomment-1564902808 if you all get some time.

shmfb:
Hello, are there any updates on this? We would also desperately wait for that feature!

PR #51, opened 2020-10-23
Fix client#removeGroupMemberships parameter order

mdgreenfield:
I wasn't able to see the following Azure error in the audit logs
`warnings` field. It looks as though even though the request to Azure
fails, when the Azure app is removed the group memberships are still
cleaned up.

I was only able to see the error below by adding a log line to print the
error when revoke is called.

```
1 error occurred:
        * error removing group membership: graphrbac.GroupsClient#RemoveMember: Failure responding to request: StatusCode=404 -- Original Error: autorest/azure: Service returned an error. Status=404 Code="Unknown" Message="Unknown service error" Details=[{"odata.error":{"code":"Request_ResourceNotFound","date":"2020-10-23T20:32:14","message":{"lang":"en","value":"Resource 'c1004989-2986-48e8-9e2e-aa03b7cacbc1' does not exist or one of its queried reference-property objects are not present."},"requestId":"a2d4d4c6-4e0a-4ea2-a807-16486c307d03"}}]
```

With that said, I think it begs the question whether or not calling `removeGroupMemberships` should even be called on revoke.


client.go
```diff
@@ -302,7 +302,7 @@ func (c *client) removeGroupMemberships(ctx context.Context, servicePrincipalObj
 	var merr *multierror.Error
 
 	for _, id := range groupIDs {
-		if _, err := c.provider.RemoveGroupMember(ctx, servicePrincipalObjectID, id); err != nil {
+		if _, err := c.provider.RemoveGroupMember(ctx, id, servicePrincipalObjectID); err != nil {
```
mdgreenfield, line 305: ref to RemoveGroupMember - https://github.com/hashicorp/vault-plugin-secrets-azure/blob/6544951c66d6c23b467877757d5f58dc7741dace/provider.go#L41

mdgreenfield:
I came across this working on an implementation of https://github.com/hashicorp/vault-plugin-secrets-azure/issues/50 with our internal fork.

austingebauer (approved):
I realize it's been a while on this one, but it makes sense to fix the parameter order. Thanks, @mdgreenfield!

PR #53, opened 2021-01-05
Added Base URI call to authorization SDK

hiscjo01:
# Overview
Changed azure-sdk-for-go implementation, passing the Resource Graph endpoint.
This adds support for Government cloud across regions and prevents a mis-match of URI when requesting an authorization token

I have tested this works with Gov Cloud (and that the Commercial implementation has now been impacted).

The plugin current uses the NewRoleAssignmentsClient & NewRoleDefinitionsClient functions.
These do not take a Resource Manager Endpoint input and use the libraries default. (https://management.azure.com)

This can be seen in the following files:

**Functions Definition**
- https://github.com/hashicorp/vault-plugin-secrets-azure/blob/master/vendor/github.com/Azure/azure-sdk-for-go/services/preview/authorization/mgmt/2018-01-01-preview/authorization/roledefinitions.go#L34
- https://github.com/hashicorp/vault-plugin-secrets-azure/blob/master/vendor/github.com/Azure/azure-sdk-for-go/services/preview/authorization/mgmt/2018-01-01-preview/authorization/roleassignments.go#L35

**Default Value:** 
- https://github.com/hashicorp/vault-plugin-secrets-azure/blob/master/vendor/github.com/Azure/azure-sdk-for-go/services/preview/authorization/mgmt/2018-01-01-preview/authorization/client.go#L29

The token that is generated using this DefaultBaseURI is invalid for use against the Government API and returns an error.


catsby (approved):
LGTM, thanks!

calvn (approved):

PR #55, opened 2021-02-05
set GUID in user-agent

catsby:
# Overview

Adds a GUID into the user-agent header, depending on the Vault version (enterprise or not)

### DIFF
--- a/provider.go
+++ b/provider.go
@@ -10,7 +10,6 @@ import (
 	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"github.com/hashicorp/vault/sdk/helper/useragent"
 	"github.com/hashicorp/vault/sdk/version"
-	"github.com/ryboe/q"
 )
 
 // AzureProvider is an interface to access underlying Azure client objects and supporting services.
@@ -105,7 +104,6 @@ func newAzureProvider(settings *clientSettings) (AzureProvider, error) {
 		vaultIDString = "; b2c13ec1-60e8-4733-9a76-88dbb2ce2471)"
 	}
 	userAgent = strings.Replace(userAgent, ")", vaultIDString, 1)
-	q.Q(userAgent)
 
 	appClient := graphrbac.NewApplicationsClient(settings.TenantID)
 	appClient.Authorizer = authorizer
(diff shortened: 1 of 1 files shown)


provider.go
```diff
@@ -85,6 +88,25 @@ func newAzureProvider(settings *clientSettings) (AzureProvider, error) {
...
+	// Vault/1.6.0 (+https://www.vaultproject.io/; azure-secrets; go1.15.7)
+	//
+	// Here we append a unique code if it's an enterprise version, where
+	// VersionMetadata will contain a non-empty string like "ent" or "prem".
+	// Otherwise use the default identifier for OSS Vault. The end result looks
+	// like so:
+	//
+	// Vault/1.6.0 (+https://www.vaultproject.io/; azure-secrets; go1.15.7; b2c13ec1-60e8-4733-9a76-88dbb2ce2471)
+	vaultIDString := "; 15cd22ce-24af-43a4-aa83-4c1a36a4b177)"
+	ver := version.GetVersion()
```
calvn, line 103: This might be not be accurate if the plugin is compiled and ran externally on a Vault Enterprise installation since this will always be pointing to the OSS SDK version value.

catsby: Unfortunately I wasn't able to find a way of getting any metadata or otherwise information on being Enterprise or not when compiling the plugin separately. Do you have any suggestions, or a better way of handling it?

provider.go
```diff
@@ -2,12 +2,15 @@ package azuresecrets
...
 	"context"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/services/graphrbac/1.6/graphrbac"
 	"github.com/Azure/azure-sdk-for-go/services/preview/authorization/mgmt/2018-01-01-preview/authorization"
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"github.com/hashicorp/vault/sdk/helper/useragent"
+	"github.com/hashicorp/vault/sdk/version"
+	"github.com/ryboe/q"
```
kalafut, line 13: Leftover debugging code.

catsby: 🤦 fixed in 272fa7a

kalafut (approved):

PR #59, opened 2021-03-16
Use the MS Graph API for atomic add/remove password operations

### LINKED ISSUE #58: Race Condition in Password Creation Results in Invalid Password
When generating Azure credentials using a role configured with an existing service principal, there is a race condition in the password creation logic which can result in a password being returned to the caller but not set on the Azure application.

The [logic is](https://github.com/hashicorp/vault-plugin-secrets-azure/blob/0677af8c393c20d47dede2e25061921ce6cfc28f/client.go#L116-L162) as follows:
- List Passwords (note that `keyId`s are returned here, no actual passwords)
- [add/remove password from list}
- Update Passwords (if a `keyId` is not present, the password is removed)

As a result, it is possible for request A and request B to retrieve the same list of existing passwords (i.e. `keyId`s) but the last request from A or B to update the passwords will win.

Although, there is [locking in place](https://github.com/hashicorp/vault-plugin-secrets-azure/blob/ac95810c9b5049bf6f49b2ccb571cf486fcb82f6/path_service_principal.go#L160-L161) to prevent that, this does not handle the case where an existing service principal is configured on multiple Azure secret engine mounts or across multiple Vault clusters.

The current logic is using the [deprecated](https://techcommunity.microsoft.com/t5/azure-active-directory-identity/update-your-applications-to-use-microsoft-authentication-library/ba-p/1257363) Azure AD Graph API, however, the newer Microsoft Graph API has endpoints to atomically [addPassword](https://docs.microsoft.com/en-us/graph/api/application-addpassword?view=graph-rest-1.0&tabs=http) and [removePassword](https://docs.microsoft.com/en-us/graph/api/application-removepassword?view=graph-rest-1.0&tabs=http).

Because the MS Graph API has a different set of permissions required for the plugin configuration, it is probably better to have a flag on the plugin configuration that defaults to the Azure AD Graph API but which allows users to upgrade to the MS Graph API once their permission changes have been added.

mdgreenfield:
Azure Active Directory Graph API, now deprecated, does not provide
support for atomically creating/removing passwords on an application. As
a result, there is a race conditions that can occur when creds are being
created for roles configured with an existing service principal that
is configured on multiple mounts or across multiple Vault clusters.

Unfortunately, [`Azure/azure-sdk-for-go`](https://github.com/Azure/azure-sdk-for-go) does not yet offer a MS Graph API 
client, therefore, this PR utilizes [`Azure/go-autorest`](https://github.com/Azure/go-autorest) to construct a client the 
same as [`Azure/azure-sdk-for-go`](https://github.com/Azure/azure-sdk-for-go).

This changeset preserves using the AAD Graph API by default but provides
a mount configuration option for toggling to the new MS Graph API. This
is because the two APIs require different API permissions. This allows
users to upgrade to the new plugin version and then switch to the new
API.

Additionally, although using the MS Graph API is a net benefit, it
itself has reliability issues when handling multiple requests in
parallel. More details can be found in
https://github.com/mdgreenfield/microsoft-graph-api-reliability and I am
working with Microsoft to try to get some of these reliability issues
resolved.

Fixes #58

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -14,7 +14,7 @@ import (
 type azureSecretBackend struct {
 	*framework.Backend
 
-	getProvider func(*clientSettings) (AzureProvider, error)
+	getProvider func(*clientSettings, bool, passwords) (AzureProvider, error)
 	client      *client
 	settings    *clientSettings
 	lock        sync.RWMutex
@@ -121,14 +121,14 @@ func (b *azureSecretBackend) getClient(ctx context.Context, s logical.Storage) (
 		b.settings = settings
 	}
 
-	p, err := b.getProvider(b.settings)
-	if err != nil {
-		return nil, err
-	}
-
 	passwords := passwords{
 		policyGenerator: b.System(),
 		policyName:      config.PasswordPolicy,
+	}
+
+	p, err := b.getProvider(b.settings, config.UseMsGraphAPI, passwords)
+	if err != nil {
+		return nil, err
 	}
 
 	c := &client{
--- a/backend_test.go
+++ b/backend_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -13,6 +14,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/graphrbac/1.6/graphrbac"
 	"github.com/Azure/azure-sdk-for-go/services/preview/authorization/mgmt/2018-01-01-preview/authorization"
 	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/date"
 	"github.com/Azure/go-autorest/autorest/to"
 	log "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/vault/sdk/helper/logging"
@@ -44,7 +46,7 @@ func getTestBackend(t *testing.T, initConfig bool) (*azureSecretBackend, logical
 
 	b.settings = new(clientSettings)
 	mockProvider := newMockProvider()
-	b.getProvider = func(s *clientSettings) (AzureProvider, error) {
+	b.getProvider = func(s *clientSettings, usMsGraphApi bool, p passwords) (AzureProvider, error) {
 		return mockProvider, nil
 	}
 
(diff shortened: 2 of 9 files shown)


mdgreenfield:
Relates to https://github.com/hashicorp/vault-plugin-secrets-azure/pull/63

mdgreenfield:
I'm happy to get this PR updated but would like some feedback from the maintainers about this approach first. #63 is a bit duplicative so we need to come to a consensus on the best way to add support for the new API. I do believe a config flag to toggle is necessary given the different permissions that are required for the new API otherwise there is a risk of breaking users simply from a Vault upgrade.

jasonodonnell:
@mdgreenfield We're looking at both of these PRs internally and should have some comments on both PRs soon. Sorry for the delay but this definitely something we're looking to change soon 😄

jasonodonnell (commented):
Hi @mdgreenfield, thanks for the contribution! I've taken a closer look at this and did some manual testing. In general this looks really good and we do agree that having the msgraph configurable is the right approach. Unfortunately during my testing I could not generate a dynamic application with a group. I've tried [various permissions](https://docs.microsoft.com/en-us/graph/api/group-list?view=graph-rest-1.0&tabs=http) but keep getting:

```
* 1 error occurred:
	* unable to lookup Azure group: graphrbac.GroupsClient#List: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code="Unknown" Message="Unknown service error" Details=[{"odata.error":{"code":"Authorization_RequestDenied","date":"2021-08-16T15:51:15","message":{"lang":"en","value":"Insufficient privileges to complete the operation."},"requestId":"37e2f04f-b0de-4bb6-8548-8c7c9885818c"}}]
```

This may be an issue on my end but I wanted to ask if you were able to get this to work and what permissions did you give the service principal? I suspect these errors are due to not updating the group code but wanted to double check.

mdgreenfield:
@jasonodonnell, do you only have Microsoft Graph permissions assigned? I think you may be running into the fact that I only implemented the client for AAD Applications so at the moment it requires both Microsoft Graph permissions and Azure Active Directory Graph permissions.

If your team is comfortable with the changes I've laid out I can probably find some time to implement the remaining functionality for Groups, RoleAssignments, ServicePrincipals, etc...

jasonodonnell:
Ah yes, that was it, I removed all AAD permissions to test just the MS graph work. I missed that the groups client wasn't updated as well. I'm meeting with my team tomorrow to discuss which approach we're moving forward with. Thanks!

jasonodonnell:
@mdgreenfield Sorry for the delay, but we've decided to move forward with this implementation since its closer to other Vault projects using MS graph.

The next steps would be:
* Resolve any conflicts in the PR
* Add MS graph support for other operations in the plugin.

If you aren't available to add extra support, we can take over the PR and add the functionality. Please let me know how you would like to proceed.

mdgreenfield:
@jasonodonnell, I should be able to get the PR rebased to resolve the conflicts today but if your team is able to run with this PR that would probably be best because I don't think I'll have the time to contribute for a few weeks.

PR #64, opened 2021-08-03
Add rotate-root endpoint

jasonodonnell:
This adds a new endpoint to support rotation of the root password in Azure. I had to rearrange some of the WAL code to support the root rotation because the current WAL rollback code deletes the service principal app when there's an error.

I'm still working on updating the test cases for the new endpoints, but the core logic is ready for review.

## Testing

Setting up an Azure root account for Vault can be difficult, but [our learn guide](https://learn.hashicorp.com/tutorials/vault/azure-secrets?in=vault/secrets-management#create-an-azure-service-principal-and-resource-group) does a good job of guiding an operator through it.

```bash
$ vault secrets enable azure

$ vault write azure/config \
        subscription_id=$SUBSCRIPTION_ID  \
        client_id=$CLIENT_ID \
        client_secret=$CLIENT_SECRET \
        tenant_id=$TENANT_ID \

$ vault write -f azure/rotate-root
Success! Data written to: azure/rotate-root
```

After rotation in Azure you can view the new secret attached to the application.

PR #70, opened 2021-10-04
Add rotate-root endpoint

pcman312:
# Overview
Adds a new endpoint (`rotate-root`) that rotates the credentials for the root user Vault uses to talk to Azure. This will delete any existing credentials as the root user should be exclusively owned by Vault after calling this endpoint.

# Design of Change
The rotation logic is a little different from our other engines in that Vault doesn't know the credential before sending the request to Azure. This is a part of the [Microsoft Graph API](https://docs.microsoft.com/en-us/graph/api/application-addpassword?view=graph-rest-1.0&tabs=http). This ends up changing the WAL code from performing a rollback to performing a cleanup. If the password creation succeeds in Azure, but the deletion of the existing passwords fails for some reason, a WAL is used to retry deleting those passwords. If any additional passwords are created after the rotate-root endpoint is hit, those passwords are not deleted. Only the passwords the existed at the time of the endpoint being hit will be deleted.

Behavioral notes:
- If a separate request happens shortly after rotate-root, Vault may receive an error from Azure indicating the credentials do not work. In my testing, the new password started working after a handful of seconds.
- Expiration date is set to 6 months past the current time. This is currently not configurable. The expiration date is returned from Vault

- Big thank you to @jasonodonnell who put together an earlier version of this: https://github.com/hashicorp/vault-plugin-secrets-azure/pull/64 I referenced this PR a lot when creating this one

# Contributor Checklist
- [ ] Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet
  TBD
- [x] Backwards compatible
- [x] Add output for any tests not ran in CI to the PR description (eg, acceptance tests)
<details>
<summary>Manual test</summary>

```
$ vault secrets enable --path azure-msgraph azure
Success! Enabled the azure secrets engine at: azure-msgraph/

$ vault write azure-msgraph/config subscription_id="${SUBSCRIPTION_ID}" client_id="${CLIENT_ID}" client_secret="${CLIENT_SECRET}" tenant_id="${TENANT_ID}" use_microsoft_graph_api=true
Success! Data written to: azure-msgraph/config

$ vault write -f -format=json azure-msgraph/rotate-root | jq
{
  "request_id": "c6f35803-f89a-e106-16a9-fcec1fb15280",
  "lease_id": "",
  "lease_duration": 0,
  "renewable": false,
  "data": {
    "display_name": "vault-plugin-secrets-azure-91be0d3c-f958-b5bb-70a7-7e96c4ccc133",
    "end_date": "2022-04-04T17:22:41.728754Z",
    "secret_id": "a2351943-a416-4dd6-9807-5d41386a75ef"
  },
  "warnings": null
}
```
</details>


path_rotate_root.go
```diff
@@ -0,0 +1,248 @@
...
+		},
+
+		HelpSynopsis: "Attempt to rotate the root credentials used to communicate with Azure.",
+		HelpDescription: "This path will attempt to generate new root credentials for the user used to access and manipulate Azure.\n" +
+			"The new credentials will not be returned from this endpoint, nor the read config endpoint.",
+	}
+}
+
+func (b *azureSecretBackend) pathRotateRoot(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
+	expiration := time.Now().AddDate(0, 6, 0)
```
jasonodonnell, line 41: Should be this be configurable?

path_rotate_root.go
```diff
@@ -0,0 +1,248 @@
...
+		credsToDelete = append(credsToDelete, *cred.KeyID)
+	}
+
+	uniqueID, err := uuid.GenerateUUID()
+	if err != nil {
+		return api.PasswordCredential{}, nil, fmt.Errorf("failed to generate UUID: %w", err)
+	}
+
+	// This could have the same username customization logic put on it if we really wanted it here
+	passwordDisplayName := fmt.Sprintf("vault-plugin-secrets-azure-%s", uniqueID)
```
jasonodonnell, line 103: Wondering if the prefix is a bit long and could simply just be `vault_%s`, similar to application passwords?

path_rotate_root.go
```diff
@@ -0,0 +1,248 @@
...
+		merr := multierror.Append(err, azureErr)
+		return api.PasswordCredential{}, nil, merr
+	}
+
+	err = removeApplicationPasswords(ctx, storage, client.provider, *app.ID, credsToDelete...)
+	if err != nil {
+		warnings = append(warnings, fmt.Sprintf("failed to clean up all other credentials for root user. Will attempt again later.\n%s", err.Error()))
+	}
+
+	return newPasswordResp.PasswordCredential, warnings, nil
```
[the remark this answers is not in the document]
pcman312: That's not a bad idea to include the warning in most/all of the other endpoints. Especially considering we don't know what changes/limitations Azure might make to AAD prior to it being removed.

path_rotate_root.go
pcman312: Good call. I'll add this to the endpoint

path_config.go
```diff
@@ -129,7 +137,20 @@ func (b *azureSecretBackend) pathConfigWrite(ctx context.Context, req *logical.R
 
 	err = b.saveConfig(ctx, config, req.Storage)
 
-	return nil, err
+	return addAADWarning(nil, config), err
+}
+
+func addAADWarning(resp *logical.Response, config *azureConfig) *logical.Response {
```
jasonodonnell, line 143: Love this

path_rotate_root.go
```diff
@@ -0,0 +1,256 @@
...
+
+	"github.com/hashicorp/vault/sdk/framework"
+	"github.com/hashicorp/vault/sdk/logical"
+)
+
+func pathRotateRoot(b *azureSecretBackend) *framework.Path {
+	return &framework.Path{
+		Pattern: "rotate-root",
+		Fields: map[string]*framework.FieldSchema{
+			"expiration": {
```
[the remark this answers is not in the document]
pcman312: I could go either way on this. I thought it made sense here since it only applies to rotating the root credentials, and nothing to do with any other part of the config. The other rotate-roots don't take any input because they don't have expiration dates like Azure does, so I'm not sure that's a good comparison. For what it's worth, the expiration here is an optional field and will default to ~6 months so you don't have to explicitly specify it unless you want a different value.

path_rotate_root.go
```diff
@@ -0,0 +1,256 @@
...
+	resultData := map[string]interface{}{
+		"secret_id": *passCred.KeyID,
+		"end_date":  *passCred.EndDate,
+	}
+	if passCred.DisplayName != nil && *passCred.DisplayName != "" {
+		resultData["display_name"] = *passCred.DisplayName
+	}
+
+	resp := &logical.Response{
+		Data:     resultData,
```
jasonodonnell, line 69: I don't see anywhere else where we return results in a rotate root. Looks like this on my end:
```
Key             Value
---             -----
display_name    vault-ae443608-693b-6804-b463-f215f09396bc
end_date        2022-04-22T20:28:19.459986Z
secret_id       0ba38415-a58d-4983-b7f1-4cb6da68080f
```

None of this seems particular useful, so I'm wondering if we should just remove the data from the payload. Thoughts?

jasonodonnell (approved):
Minor comments but overall this tested well and looks good to me!

path_rotate_root.go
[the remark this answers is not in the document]
jasonodonnell: My vote is in `config` with other root related values, so an operator can configure once and not worry about it anymore.

pcman312:
Thanks @mdgreenfield !

> My recommendation here would be to continue to use the existing credentials (if valid) for X. Where X could be either a duration or some heuristic to try and determine when the new credentials are valid and consistent.

I'm brainstorming how we could do this effectively. I'm thinking we could use the WAL to save the config & remove old credentials instead of just removing the old credentials. This would result in a new credential being created immediately with the internal client would continue to use the old credential. The WAL would trigger after ~10 minutes which would then save the new configuration and remove the old credentials. This would result in the operations to continue to use the old credential for ~10 minutes.

path_rotate_root.go
```diff
@@ -0,0 +1,256 @@
...
+)
+
+func pathRotateRoot(b *azureSecretBackend) *framework.Path {
+	return &framework.Path{
+		Pattern: "rotate-root",
+		Fields: map[string]*framework.FieldSchema{
+			"expiration": {
+				Type: framework.TypeString,
+				// 28 weeks (~6 months) -> days -> hours
+				Default:     (28 * 7 * 24 * time.Hour).String(),
```
calvn, line 24: How was the default value of 6 months chosen here? Would it be good to go with [Azure's default value](https://docs.microsoft.com/en-us/graph/api/application-addpassword?view=graph-rest-1.0&tabs=http#request-body) in here?

path_rotate_root.go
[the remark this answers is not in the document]
calvn: Since we already provide the root's client ID and client secret in the config endpoint, having this in the config endpoint as well seems reasonable. I don't think that this value will change once it's set in there, and I'm not sure if operators would set a different expiration value from one rotation to the other. A side benefit of having it in the config is that it can be read back when one is `GET`'ing the config.

path_rotate_root.go
```diff
@@ -0,0 +1,256 @@
...
+	aMap := map[string]struct{}{}
+	for _, aStr := range a {
+		aMap[aStr] = struct{}{}
+	}
+
+	result := []string{}
+	for _, bStr := range b {
+		if _, exists := aMap[bStr]; exists {
+			result = append(result, bStr)
+			continue
```
calvn, line 252: Is `continue` needed here?

path_rotate_root.go
pcman312: It's the default when creating a credential in the Azure UI. I have no idea why it's different from the API's default.

api/application_msgraph.go
```diff
@@ -71,6 +71,30 @@ func (c *AppClient) GetApplication(ctx context.Context, applicationObjectID stri
 	return result, nil
 }
 
+type listApplicationsResponse struct {
+	Value []ApplicationResult
```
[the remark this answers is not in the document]
pcman312: Good catch. Added

path_config.go
```diff
@@ -64,12 +68,27 @@ func pathConfig(b *azureSecretBackend) *framework.Path {
 				Type:        framework.TypeBool,
 				Description: "Enable usage of the Microsoft Graph API over the deprecated Azure AD Graph API.",
 			},
+			"default_expiration": &framework.FieldSchema{
+				Type: framework.TypeString,
```
kalafut, line 72: Can this be `framework.TypeDurationSecond` and let the framework handle the parsing?

path_config.go
```diff
@@ -64,12 +68,27 @@ func pathConfig(b *azureSecretBackend) *framework.Path {
 				Type:        framework.TypeBool,
 				Description: "Enable usage of the Microsoft Graph API over the deprecated Azure AD Graph API.",
 			},
+			"default_expiration": &framework.FieldSchema{
```
kalafut, line 71: Seeing "default" here, I expected that the rotate root could optionally accept an expiration (and we'd use the default if not provided). But there isn't any expiration parameter during rotation. Should there be? If not, then perhaps this shouldn't be called "default".

path_rotate_root.go
```diff
@@ -0,0 +1,244 @@
...
+	"github.com/hashicorp/vault/sdk/logical"
+	"github.com/mitchellh/mapstructure"
+)
+
+func pathRotateRoot(b *azureSecretBackend) *framework.Path {
+	return &framework.Path{
+		Pattern: "rotate-root",
+		Fields:  map[string]*framework.FieldSchema{
+			// None
+		},
```
kalafut, line 21:
```suggestion
```

path_config.go
```diff
@@ -64,12 +68,27 @@ func pathConfig(b *azureSecretBackend) *framework.Path {
...
+				Type: framework.TypeString,
+				// 28 weeks (~6 months) -> days -> hours
+				Default:     (28 * 7 * 24 * time.Hour).String(),
```
kalafut, line 74: Preferable to define this duration as a constant as it is used in multiple place. A comment about how 6 months was arrived at would also be good.

api/application_msgraph.go
```diff
@@ -108,23 +133,19 @@ func (c *AppClient) DeleteApplication(ctx context.Context, applicationObjectID s
...
 	}
 
-	err = autorest.Respond(
+	return autorest.Respond(
 		resp,
 		c.client.ByInspecting(),
 		azure.WithErrorUnlessStatusCode(http.StatusNoContent, http.StatusNotFound),
 		autorest.ByClosing())
-
 	if err != nil {
```
austingebauer, line 141: This code is unreachable with the return on L136.

path_rotate_root.go
```diff
@@ -0,0 +1,241 @@
...
+
+	// This could have the same username customization logic put on it if we really wanted it here
+	passwordDisplayName := fmt.Sprintf("vault-%s", uniqueID)
+
+	newPasswordResp, err := client.provider.AddApplicationPassword(ctx, *app.ID, passwordDisplayName, expiration)
+	if err != nil {
+		return api.PasswordCredential{}, fmt.Errorf("failed to add new password: %w", err)
+	}
+
+	// Write a WAL with the new credential (and some other info) to write to the config later
```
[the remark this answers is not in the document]
jasonodonnell: I believe it's done this way because `AddApplicationPassword` also generates the password and returns it. We can then use that password in the WAL.

api/application_msgraph.go
jasonodonnell: This was due to a bad rebase. Good catch!

api/application_msgraph.go
```diff
@@ -72,6 +72,30 @@ func (c *AppClient) GetApplication(ctx context.Context, applicationObjectID stri
 	return result, nil
 }
 
+type listApplicationsResponse struct {
```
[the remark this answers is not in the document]
jasonodonnell: Good catch! This is fixed.

path_rotate_root.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	expiration := time.Now().Add(expDur)
+
+	passCred, err := b.rotateRootCredentials(ctx, req.Storage, *config, expiration)
+	if err != nil {
+		return nil, err
+	}
+
+	resultData := map[string]interface{}{
+		"secret_id": *passCred.KeyID,
```
calvn, line 51: If we don't want to return this date on rotate-root, maybe we can move it to the config read endpoint.

path_config.go
```diff
@@ -123,13 +143,39 @@ func (b *azureSecretBackend) pathConfigWrite(ctx context.Context, req *logical.R
...
+const aadWarning = "This configuration is using the Azure Active Directory API which is being " +
+	"removed soon. Please migrate to using the Microsoft Graph API using the " +
+	"use_microsoft_graph_api configuration parameter."
+
+func addAADWarning(resp *logical.Response, config *azureConfig) *logical.Response {
+	if config.UseMsGraphAPI {
+		return resp
+	}
+	if resp == nil {
+		resp = &logical.Response{}
```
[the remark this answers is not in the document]
calvn: Oh right!

path_config.go
```diff
@@ -64,12 +70,26 @@ func pathConfig(b *azureSecretBackend) *framework.Path {
 				Type:        framework.TypeBool,
 				Description: "Enable usage of the Microsoft Graph API over the deprecated Azure AD Graph API.",
 			},
+			"root_password_expiration": &framework.FieldSchema{
```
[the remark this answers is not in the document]
calvn: I'm fine with keeping it as-is, but it might help to change the description to `The expiration TTL of the new credentials in Azure.` instead.

PR #75, opened 2021-10-29
Fix segfault when config is nil

jasonodonnell:
If `config` is nil (backend is enabled but not yet configured), the periodic function will fail because it does not check if the config is nil. Added a guard and a test to cover this case.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -195,15 +195,15 @@ func (b *azureSecretBackend) getClient(ctx context.Context, s logical.Storage) (
 			config = new(azureConfig)
 		}
 
-		if config == nil {
-			return nil, fmt.Errorf("config is nil")
-		}
-
 		settings, err := b.getClientSettings(ctx, config)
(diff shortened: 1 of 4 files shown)


vinay-gopalan (approved):

backend.go
```diff
@@ -184,6 +190,10 @@ func (b *azureSecretBackend) getClient(ctx context.Context, s logical.Storage) (
 		return nil, err
 	}
 
+	if config == nil {
+		return nil, fmt.Errorf("config is nil")
+	}
+
 	if b.settings == nil {
 		if config == nil {
 			config = new(azureConfig)
```
tvoran, line 199: Looks like this block will never be triggered if we're guarding against `config == nil` on line 193. 

Perhaps we could move the check on line 193 down below the `if b.settings == nil` block?

tvoran (approved):

PR #81, opened 2021-11-12
Add support for testing enterprise.

benashz:
Add enterprise support to acceptance tests.

### DIFF
--- a/tests/acceptance/basic.bats
+++ b/tests/acceptance/basic.bats
@@ -61,6 +61,7 @@ HERE
       -e VAULT_DEV_ROOT_TOKEN_ID="${VAULT_TOKEN?}" \
       -e VAULT_ADDR="http://localhost:8200" \
       -e VAULT_DEV_LISTEN_ADDRESS="0.0.0.0:8200" \
+      -e VAULT_LICENSE="${VAULT_LICENSE}" \
       --privileged \
       --detach \
       ${VAULT_IMAGE?}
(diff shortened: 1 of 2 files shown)


tests/acceptance/basic.bats
```diff
@@ -61,6 +61,7 @@ HERE
       -e VAULT_DEV_ROOT_TOKEN_ID="${VAULT_TOKEN?}" \
       -e VAULT_ADDR="http://localhost:8200" \
       -e VAULT_DEV_LISTEN_ADDRESS="0.0.0.0:8200" \
+      -e VAULT_LICENSE="${VAULT_LICENSE}" \
```
tomhjp, line 64: Does this fail in a quick and obvious way if VAULT_LICENSE isn't set? If not, could we add a little check that prints a nice error message? Or if you want it to be able to run with or without, you could just do `-e VAULT_LICENSE` and docker will pull it through for you.

benashz: It all depends on the docker image you use. The tests default to OSS, so setting a license is not required. In order to test with enterprise you would have to also set `VAULT_LICENSE`, and `VAULT_IMAGE` to an enterprise image  e.g. `VAULT_IMAGE=hashicorp/vault-enterprise:1.9.0-rc1_ent`. Vault Enterprise will obviously fail to start if a valid license is not provided.

I am hesitant to add code that parses the docker image name. However adding something to docs would probably be good.

tomhjp: Makes sense 👍 yeah +1 it would be great to have a little bit of documentation somewhere to help people test enterprise, even if it's just code comments.

tomhjp (approved):

PR #85, opened 2021-11-25
Fix bats based MS Graph acceptance test

benashz:
The permissions for the MS Graph tests were incomplete, unfortunately
the issue was not caught by the bats test. Essentially doing a local
variable assignment from a sub-shell's stdout does not result in the
test failing when the sub-shell exits with a non-zero status

Summary of fixes:
- bats: add support testing Azure data plane roles
- terraform: ensure that 'MS Graph' test has all required permissions for
  the plugin to generate secrets
- terraform: wait for grant assignment after `admin-consent`
- bats: factor out common code to `common.sh`
- bats: ensure that all variable assignment from sub-shell values fail the
  test in the case that the sub-shell process exits with a non-zero
  status
- bats: add better test logging
- bats: add support for specifying a custom log file
- build: optionally build the plugin for the target vault docker image
  os/arch
- build: extend to better support bats acceptance test from make

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,13 @@ EXTERNAL_TOOLS=\
 BUILD_TAGS?=${TOOL}
 GOFMT_FILES?=$$(find . -name '*.go' | grep -v vendor)
 
+# Acceptance test variables
+WITH_DEV_PLUGIN?=
+AZURE_TENANT_ID?=
+SKIP_TEARDOWN?=
+TESTS_OUT_FILE?=
+TESTS_FILTER?='.*'
+
 # bin generates the releaseable binaries for this plugin
 bin: fmtcheck generate
 	@CGO_ENABLED=0 BUILD_TAGS='$(BUILD_TAGS)' sh -c "'$(CURDIR)/scripts/build.sh'"
@@ -22,6 +29,8 @@ dev: fmtcheck generate
 	@CGO_ENABLED=0 BUILD_TAGS='$(BUILD_TAGS)' VAULT_DEV_BUILD=1 sh -c "'$(CURDIR)/scripts/build.sh'"
 dev-dynamic: generate
 	@CGO_ENABLED=1 BUILD_TAGS='$(BUILD_TAGS)' VAULT_DEV_BUILD=1 sh -c "'$(CURDIR)/scripts/build.sh'"
+dev-acceptance: fmtcheck generate
+	@CGO_ENABLED=0 BUILD_TAGS='$(BUILD_TAGS)' VAULT_DEV_BUILD= XC_OSARCH=linux/amd64 sh -c "'$(CURDIR)/scripts/build.sh'"
 
 testcompile: fmtcheck generate
 	@for pkg in $(TEST) ; do \
@@ -37,8 +46,8 @@ test: fmtcheck generate
 	VAULT_ACC=1 go test -tags='$(BUILD_TAGS)' $(TEST) -v $(TESTARGS) -timeout 45m
 
 # test-acceptance runs all acceptance tests
-test-acceptance:
-	bats $(CURDIR)/tests/acceptance/basic.bats
+test-acceptance: $(if $(WITH_DEV_PLUGIN), dev-acceptance)
+	bats -f $(TESTS_FILTER) $(CURDIR)/tests/acceptance/basic.bats
 
 # generate runs `go generate` to build the dynamically generated
 # source files.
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -37,8 +46,8 @@ test: fmtcheck generate
 	VAULT_ACC=1 go test -tags='$(BUILD_TAGS)' $(TEST) -v $(TESTARGS) -timeout 45m
 
 # test-acceptance runs all acceptance tests
-test-acceptance:
-	bats $(CURDIR)/tests/acceptance/basic.bats
+test-acceptance: $(if $(WITH_DEV_PLUGIN), dev-acceptance)
+	bats -f $(TESTS_FILTER) $(CURDIR)/tests/acceptance/basic.bats
```
tvoran, line 50: Should we document the `WITH_DEV_PLUGIN` and `TESTS_FILTER` options in the README?

tests/acceptance/common.sh
```diff
@@ -0,0 +1,120 @@
...
+    local secret_file="${config_dir}/secret-${vault_role_name}.json"
+    vault read "${engine_name}/creds/${vault_role_name}" -format=json > ${secret_file}
+
+    sp_id=$(cat "${secret_file}" | jq -er .data.client_id)
+    log "Secret created successfully, azure_role='${role_name}', vault_role='${vault_role_name}', file=${secret_file}"
+
+    local found=''
+    for n in {0..30}
+    do
+         if ! az ad sp show --id "${sp_id}" > /dev/null ; then
```
tvoran, line 42: Is it possible to use bats primitives in this function? Like `run` and then checking `$status`? Just thinking it might be easier to read and would take care of the output without redirecting to /dev/null.

benashz: I am not sure `run` fits that well in this context. I could perhaps factor this portion out to its own function which may help with the readability issue that you mention.

Makefile
benashz: Fixed in https://github.com/hashicorp/vault-plugin-secrets-azure/pull/85/commits/322a5da4372af9d3a6295372ed241f6ddebcb660

tests/acceptance/common.sh
benashz: Fixed in https://github.com/hashicorp/vault-plugin-secrets-azure/pull/85/commits/322a5da4372af9d3a6295372ed241f6ddebcb660

tests/acceptance/common.sh
```diff
@@ -0,0 +1,138 @@
...
+function assertSPExistence() {
+    local sp_id="${1}"
+    local found=''
+    for n in {0..30}
+    do
+         if ! az ad sp show --id "${sp_id}" >&1 /dev/null ; then
+            logWarn "Failed to check service principal exists for ID ${sp_id}"
+            sleep 1
+            continue
+         fi
```
tvoran, line 79: Doesn't this need another redirect to send stdout to /dev/null?

```suggestion
         if ! az ad sp show --id "${sp_id}" >&1 > /dev/null ; then
            logWarn "Failed to check service principal exists for ID ${sp_id}"
            sleep 1
            continue
         fi
```

My earlier suggestion around using bats' `run` could look something like this too, since run always returns 0:

```suggestion
         run az ad sp show --id "${sp_id}"
         if [ $status -ne 0 ] ; then
            logWarn "Failed to check service principal exists for ID ${sp_id}: $output"
            sleep 1
            continue
         fi
```

That strikes me as being a little closer to the style you'd see in a non-bash test, which is what I was thinking as far as being more readable. YMMV.

tests/acceptance/common.sh
```diff
@@ -0,0 +1,138 @@
...
+EOF
+
+    local secret_file="${config_dir}/secret-${vault_role_name}.json"
+    vault read "${engine_name}/creds/${vault_role_name}" -format=json > ${secret_file}
+
+    sp_id=$(cat "${secret_file}" | jq -er .data.client_id)
+    log "Secret created successfully, azure_role='${role_name}', vault_role='${vault_role_name}', file=${secret_file}"
+
+    if !assertSPExistence ; then
+        exit 1
```
tvoran, line 40: Was a little surprised this is `exit` instead of `return`, but is that because each `@test` is running in a subshell?

tvoran (approved):
Left a few questions, but nothing blocking as far as I'm concerned.

tests/acceptance/common.sh
benashz: Fixed in https://github.com/hashicorp/vault-plugin-secrets-azure/pull/85/commits/de8335c6c878e2d9b19e4dcfaaefb6038dd6aa29

tests/acceptance/common.sh
benashz: Definitely a good point, but I feel that these functions are common to the tests and could be used outside of bats, so I think I would prefer to keep it as-is.

PR #87, opened 2022-01-31
Fix bug where graph API didn't change when environment is set

jasonodonnell:
The graph URL was hardcoded to `https://graph.microsoft.com` in the new Graph API, but the URL changes depending on the type of Azure cloud being used (set by `environment` in the config). This adds a new helper function to switch based on which cloud the user has configured.

### DIFF
--- a/api/application_msgraph.go
+++ b/api/application_msgraph.go
@@ -15,21 +15,33 @@ import (
 	"github.com/hashicorp/go-multierror"
 )
 
-const (
-	// DefaultGraphMicrosoftComURI is the default URI used for the service MS Graph API
-	DefaultGraphMicrosoftComURI = "https://graph.microsoft.com"
-)
-
 var _ ApplicationsClient = (*AppClient)(nil)
 var _ GroupsClient = (*AppClient)(nil)
 var _ ServicePrincipalClient = (*AppClient)(nil)
 
 type AppClient struct {
-	client authorization.BaseClient
+	client   authorization.BaseClient
+	graphURI string
 }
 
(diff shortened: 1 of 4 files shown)


api/application_msgraph.go
```diff
@@ -15,21 +15,33 @@ import (
...
 type AppClient struct {
-	client authorization.BaseClient
+	client   authorization.BaseClient
+	graphURI string
+}
+
+// Reference: https://docs.microsoft.com/en-us/graph/deployments#microsoft-graph-and-graph-explorer-service-root-endpoints
+func GetGraphURI(env string) (string, error) {
+	switch env {
+	case "AzurePublicCloud":
```
calvn, line 30: I wonder if we should default empty string to public cloud instead of error, if that happens to be the case for whatever reason.
```suggestion
	case "AzurePublicCloud", "":
```

calvn (approved):
One comment, otherwise looks good!

calvn:
This might be a good opportunity to include a changelog to the repo :)

austingebauer (approved):
LGTM. Thanks, @jasonodonnell!

api/application_msgraph.go
```diff
@@ -15,21 +15,33 @@ import (
...
+	graphURI string
+}
+
+// Reference: https://docs.microsoft.com/en-us/graph/deployments#microsoft-graph-and-graph-explorer-service-root-endpoints
+func GetGraphURI(env string) (string, error) {
+	switch env {
+	case "AzurePublicCloud", "":
+		return "https://graph.microsoft.com/", nil
+	case "AzureUSGovernmentCloud":
+		return "https://graph.microsoft.us/", nil
```
kalafut, line 33: These two URLs have a trailing slash but the others don't. Intentional?

PR #95, opened 2022-07-05
remove partner ID

Zlaticanin:
# Overview
Removing the partner ID from user-agent string in Azure secrets plugin because Microsoft is unable to track these ID's. This change doesn't affect user experience at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
 ## Next
 
+* Remove partner ID from user agent string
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
 ## Next
 
+* Remove partner ID from user agent string
```
austingebauer, line 3: Nit: We usually add a link to the PR at the end of the line (similar to below :))

austingebauer (approved):
LGTM. Thanks, @Zlaticanin!

CHANGELOG.md
Zlaticanin: Just added it

PR #98, opened 2022-08-02
Allows for persisting an app for the lifetime of the role

nbutton23:
# Overview
_A high level description of the contribution, including:_
Adds the ability to persist the created application between getting credentials.

_Who the change affects or is for (stakeholders)?_

Consumers of role definitions who may need to persist Service Principal application objects.

_What is the change?_ 

Adds a new field to the role schema that allows the application to persist for the lifetime of the role.

_Why is the change needed?_
 
Allows for the SPN to maintain ownership of resources created by it between runs.  This will be useful for another change we would like to make to allow SPNs to be assigned graphAPI permissions. Many of those permissions have something similar to the `Application.ReadWrite.OwnerBy` For this permission to work correctly we need to persist the app so that it still owns the resource. 
 
_How does this change affect the user experience (if at all)?_

No affect for existing roles, or roles that don’t use the new field.

# Design of Change
_How was this change implemented?_

On role creation, if `persist_app` is set to true, we then create a new app and bind it to the Azure roles and groups listed in the role.  We then set the `application_object_id` on the role so that we are able to reuse the `createStaticSPSecret` as the lifecycle of the credentials is the same as a static app.  

On role update we remove all Azure roles and group bindings, then create the new Azure role and group bindings. 

On role delete we delete all Azure role bindings and group membership (best effort) then the application. 

# Contributor Checklist
[My Docs PR Link](https://github.com/hashicorp/vault/pull/16537)
[ ] Backwards compatible

### DIFF
--- a/path_roles.go
+++ b/path_roles.go
@@ -339,20 +339,14 @@ func (b *azureSecretBackend) pathRoleUpdate(ctx context.Context, req *logical.Re
 }
 
 func (b *azureSecretBackend) createPersistedApp(ctx context.Context, req *logical.Request, role *roleEntry, name string) error {
+
 	c, err := b.getClient(ctx, req.Storage)
 	if err != nil {
 		return err
 	}
 
 	if role.ManagedApplicationObjectID != "" {
-		// Unassign roles
-		if err := c.unassignRoles(ctx, role.RoleAssignmentIDs); err != nil {
-			return err
-		}
-		// Removing group membership
-		if err := c.removeGroupMemberships(ctx, role.ServicePrincipalObjectID, role.GroupMembershipIDs); err != nil {
-			return err
-		}
+		removeRolesAndGroupMembership(ctx, c, role)
 
 		spObjID := role.ServicePrincipalObjectID
 
@@ -465,6 +459,8 @@ func (b *azureSecretBackend) pathRoleList(ctx context.Context, req *logical.Requ
 }
 
 func (b *azureSecretBackend) pathRoleDelete(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
+	resp := new(logical.Response)
+
 	name := d.Get("name").(string)
 	role, err := getRole(ctx, name, req.Storage)
 	if err != nil {
@@ -477,11 +473,12 @@ func (b *azureSecretBackend) pathRoleDelete(ctx context.Context, req *logical.Re
 			return nil, fmt.Errorf("error during delete: %w", err)
 		}
 
-		// unassigning roles is effectively a garbage collection operation.
-		c.unassignRoles(ctx, role.RoleAssignmentIDs)
-
-		// removing group membership is effectively a garbage collection operation.
-		c.removeGroupMemberships(ctx, role.ServicePrincipalObjectID, role.GroupMembershipIDs)
+		// unassigning roles and removing group membership is effectively a garbage collection operation.
+		// Errors will be noted but won't fail the revocation process. 
+		// Deleting the app, however, *is* required to consider the secret revoked.
+		if err := removeRolesAndGroupMembership(ctx, c, role); err != nil {
+			resp.AddWarning(err.Error())
+		}
 
 		if err = c.deleteApp(ctx, role.ApplicationObjectID, role.PermanentlyDelete); err != nil {
 			return nil, fmt.Errorf("error deleting persisted app: %w", err)
@@ -493,7 +490,7 @@ func (b *azureSecretBackend) pathRoleDelete(ctx context.Context, req *logical.Re
 		return nil, fmt.Errorf("error deleting role: %w", err)
 	}
 
-	return nil, nil
+	return resp, nil
 }
 
 func (b *azureSecretBackend) pathRoleExistenceCheck(ctx context.Context, req *logical.Request, d *framework.FieldData) (bool, error) {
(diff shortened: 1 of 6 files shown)


path_roles.go
```diff
@@ -307,6 +338,88 @@ func (b *azureSecretBackend) pathRoleUpdate(ctx context.Context, req *logical.Re
...
 
+func (b *azureSecretBackend) createPersistedApp(ctx context.Context, req *logical.Request, role *roleEntry, name string) error {
+	c, err := b.getClient(ctx, req.Storage)
+	if err != nil {
+		return err
+	}
+
+	if role.ManagedApplicationObjectID != "" {
+		// Unassign roles
+		if err := c.unassignRoles(ctx, role.RaIDs); err != nil {
```
vinay-gopalan, line 349: nit: might be worth factoring common functions for unassigning roles + removing group memberships and assigning roles + adding group memberships. Might resolve some duplicated code in this function as well as `pathRoleDelete`

vinay-gopalan (commented):
Thanks for submitting this feature! The PR's looking great 😄 
Had some initial nits/comments

path_roles.go
```diff
@@ -31,6 +31,13 @@ type roleEntry struct {
 	TTL                 time.Duration `json:"ttl"`
 	MaxTTL              time.Duration `json:"max_ttl"`
 	PermanentlyDelete   bool          `json:"permanently_delete"`
+	PersistApp          bool          `json:"persist_app"`
+
+	// Info for persisted apps
+	RaIDs                      []string `json:"role_assignment_ids"`
```
vinay-gopalan, line 37: Missed this one in the last review, sorry!

```suggestion
	RoleAssignmentIDs                      []string `json:"role_assignment_ids"`
```

path_roles.go
```diff
@@ -352,8 +466,29 @@ func (b *azureSecretBackend) pathRoleList(ctx context.Context, req *logical.Requ
...
+	}
+
+	if role != nil && role.PersistApp {
+		c, err := b.getClient(ctx, req.Storage)
+		if err != nil {
+			return nil, fmt.Errorf("error during delete: %w", err)
+		}
+
+		// unassigning roles is effectively a garbage collection operation.
+		c.unassignRoles(ctx, role.RaIDs)
```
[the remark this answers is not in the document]
nbutton23: I could do what is done https://github.com/hashicorp/vault-plugin-secrets-azure/blob/2567884f30722190dbdcf5031352c58046a14bee/path_service_principal.go#L252 and return a warning in the response. I don't know what the behavior of returning a response from this function since it currently returns nil.

path_roles.go
```diff
@@ -351,14 +459,38 @@ func (b *azureSecretBackend) pathRoleList(ctx context.Context, req *logical.Requ
...
+		c, err := b.getClient(ctx, req.Storage)
+		if err != nil {
+			return nil, fmt.Errorf("error during delete: %w", err)
+		}
+
+		// unassigning roles and removing group membership is effectively a garbage collection operation.
+		// Errors will be noted but won't fail the revocation process. 
+		// Deleting the app, however, *is* required to consider the secret revoked.
+		if err := removeRolesAndGroupMembership(ctx, c, role); err != nil {
+			resp.AddWarning(err.Error())
```
vinay-gopalan, line 480: We want to adhere to the style of returning any errors during execution, and we also want to return `nil` as the response from the Delete function instead of an empty `logical.Response` object even when everything executes correctly. In this case I would suggest changing this to be similar to what you have for the other errors above:

```suggestion
			return nil, fmt.Errorf("error during role and group membership cleanup, err=%w", err)
```

path_roles.go
```diff
@@ -351,14 +459,38 @@ func (b *azureSecretBackend) pathRoleList(ctx context.Context, req *logical.Requ
 }
 
 func (b *azureSecretBackend) pathRoleDelete(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
+	resp := new(logical.Response)
```
vinay-gopalan, line 462: If we make the changes suggested below, we won't need this object anymore

```suggestion

```

path_roles.go
```diff
@@ -351,14 +459,38 @@ func (b *azureSecretBackend) pathRoleList(ctx context.Context, req *logical.Requ
...
+	}
+
+	if role != nil && role.PersistApp {
+		c, err := b.getClient(ctx, req.Storage)
+		if err != nil {
+			return nil, fmt.Errorf("error during delete: %w", err)
+		}
+
+		// unassigning roles and removing group membership is effectively a garbage collection operation.
+		// Errors will be noted but won't fail the revocation process.
```
vinay-gopalan, line 477:
```suggestion
		// Errors will be noted but won't fail the revocation process.
```

path_roles.go
nbutton23: So removing roles and group membership is a nice to have, but isn't required, only the deletion of the app.  This follows the way this is currently done in the [none persisted app](https://github.com/hashicorp/vault-plugin-secrets-azure/blob/7f4e89517c103485b29d75ca7906fb3e6652f5ee/path_service_principal.go#L250).  As long as the app is deleted these bindings don't matter.  I can still return the error, just want to call out that it might result in deletes faililng more often since we are making more request to Azure.

vinay-gopalan (commented):
@nbutton23 sorry for the delay on this! Wanted to test the behavior of returning a response instead of `nil` more closely and wanted to confirm there were no caveats to address. I think there's one last quick fix to be made around where we initialize the response so that we still return `nil` in the case that there are no warnings. After that fix however, things are looking good and should be good to go. Thanks for your patience and time while working on this!

vinay-gopalan (approved):
Looks good! Thanks for your contribution to HashiCorp!

PR #101, opened 2022-08-22
Remove deprecated AAD graph code

Zlaticanin:
Removing AAD graph code from the Azure secrets engine. All documentation is also updated and it refers to usage of MS graph instead of AAD graph. 

PR for updating the docs: https://github.com/hashicorp/vault/pull/16819 

** Please let me know if I missed something or removed something that's still supposed to be there.

Results from running acceptance tests:
Screen Shot 2022-08-22 at 8 59 36 AM
Screen Shot 2022-08-22 at 9 00 08 AM
Screen Shot 2022-08-22 at 9 00 23 AM
Screen Shot 2022-08-22 at 9 00 36 AM


fairclothjm (commented):
I found some references to AAD in comments. Some of these are probably ok to keep around. But maybe some of them need updating?

```
$ rg -in --no-heading '^//' | rg -i aad
api/application_msgraph.go:405:// slightly different from the AAD implementation and there should be an abstraction
provider_mock_test.go:180:// AddGroupMember adds a member to a AAD Group.
provider_mock_test.go:185:// RemoveGroupMember removes a member from a AAD Group.
provider.go:187:// AddGroupMember adds a member to a AAD Group.
provider.go:192:// RemoveGroupMember removes a member from a AAD Group.
path_service_principal_test.go:488:// The credentials provided to this must include permissions to use MS Graph and not AAD
```

Maybe the comments in provider.go and provider_mock_test.go need updating?

I also found some references in the `tests` directory:

```
$ rg -in --no-heading aad tests/
tests/acceptance/basic.bats:140:@test "Azure Secrets Engine - Legacy AAD" {
tests/acceptance/basic.bats:142:    terraformInitApply ${CONFIG_DIR} -var=legacy_aad_resource_access=true
tests/acceptance/terraform/main.tf:45:variable "legacy_aad_resource_access" {
tests/acceptance/terraform/main.tf:79:    for_each = var.legacy_aad_resource_access ? [] : [""]
tests/acceptance/terraform/main.tf:110:  # Legacy Azure Active Directory Graph (AADG)
tests/acceptance/terraform/main.tf:112:    for_each = var.legacy_aad_resource_access ? [""] : []
```

Should some of these be removed?

Zlaticanin:
> I found some references to AAD in comments. Some of these are probably ok to keep around. But maybe some of them need updating?
> 
> ```
> $ rg -in --no-heading '^//' | rg -i aad
> api/application_msgraph.go:405:// slightly different from the AAD implementation and there should be an abstraction
> provider_mock_test.go:180:// AddGroupMember adds a member to a AAD Group.
> provider_mock_test.go:185:// RemoveGroupMember removes a member from a AAD Group.
> provider.go:187:// AddGroupMember adds a member to a AAD Group.
> provider.go:192:// RemoveGroupMember removes a member from a AAD Group.
> path_service_principal_test.go:488:// The credentials provided to this must include permissions to use MS Graph and not AAD
> ```
> 
> Maybe the comments in provider.go and provider_mock_test.go need updating?
> 
> I also found some references in the `tests` directory:
> 
> ```
> $ rg -in --no-heading aad tests/
> tests/acceptance/basic.bats:140:@test "Azure Secrets Engine - Legacy AAD" {
> tests/acceptance/basic.bats:142:    terraformInitApply ${CONFIG_DIR} -var=legacy_aad_resource_access=true
> tests/acceptance/terraform/main.tf:45:variable "legacy_aad_resource_access" {
> tests/acceptance/terraform/main.tf:79:    for_each = var.legacy_aad_resource_access ? [] : [""]
> tests/acceptance/terraform/main.tf:110:  # Legacy Azure Active Directory Graph (AADG)
> tests/acceptance/terraform/main.tf:112:    for_each = var.legacy_aad_resource_access ? [""] : []
> ```
> 
> Should some of these be removed?

Thank you so much JM. I updated the comments and deleted the tests that were there for AAD graph.

fairclothjm (approved):
LGTM!

One more question I have is do we plan to keep the [use_microsoft_graph_api](https://github.com/hashicorp/vault-plugin-secrets-azure/blob/2ab6ddcac1757eb756d2776da3911a64d6206b70/path_config.go#L75) config field? I see that it defaults to true so we [always set the MS Graph client](https://github.com/hashicorp/vault-plugin-secrets-azure/blob/2ab6ddcac1757eb756d2776da3911a64d6206b70/provider.go#L40). Or does it make sense to deprecate that field now?

Zlaticanin:
> LGTM!
> 
> One more question I have is do we plan to keep the [use_microsoft_graph_api](https://github.com/hashicorp/vault-plugin-secrets-azure/blob/2ab6ddcac1757eb756d2776da3911a64d6206b70/path_config.go#L75) config field? I see that it defaults to true so we [always set the MS Graph client](https://github.com/hashicorp/vault-plugin-secrets-azure/blob/2ab6ddcac1757eb756d2776da3911a64d6206b70/provider.go#L40). Or does it make sense to deprecate that field now?

Great point! I think that it makes sense to deprecate the config field for MS Graph as well, since it's always true now. I am going to wait for one more review, and if they agree I will go ahead and remove that part of the code too. 
Thank you JM!

fairclothjm:
@MilenaHC 
> I will go ahead and remove that part of the code too.

I am not sure removing it right out is the right course of action. Maybe we should make a deprecation notice in the [docs](https://www.vaultproject.io/docs/deprecation) first? Then possibly a changelog entry in vault with the `deprecation` type. Also `framework.Backend` has a `Deprecated` field that we could set for the field in path_config.go but I don't know how common it is to use that.

path_config.go
```diff
@@ -167,24 +167,7 @@ func (b *azureSecretBackend) pathConfigWrite(ctx context.Context, req *logical.R
...
-func addAADWarning(resp *logical.Response, config *azureConfig) *logical.Response {
-	if config.UseMsGraphAPI {
-		return resp
-	}
-	if resp == nil {
-		resp = &logical.Response{}
-	}
-	resp.AddWarning(aadWarning)
-	return resp
+	return nil, err
```
austingebauer, line 170: We'll want to remove the ability to set `use_microsoft_graph_api` in [`pathConfigWrite`](https://github.com/hashicorp/vault-plugin-secrets-azure/blob/main/path_config.go#L147-L151). Right now if you set its value to false, the plugin will panic when trying to read credentials because the MS graph API client is nil.

I'd also say we can completely remove [the field](https://github.com/hashicorp/vault-plugin-secrets-azure/blob/main/path_config.go#L37) from `azureConfig`. This will also mean removing its usage from other locations in the code.

path_config.go
```diff
@@ -200,20 +183,19 @@ func (b *azureSecretBackend) pathConfigRead(ctx context.Context, req *logical.Re
...
-			"tenant_id":               config.TenantID,
-			"environment":             config.Environment,
-			"client_id":               config.ClientID,
-			"use_microsoft_graph_api": config.UseMsGraphAPI,
-			"root_password_ttl":       int(config.RootPasswordTTL.Seconds()),
+			"subscription_id":   config.SubscriptionID,
+			"tenant_id":         config.TenantID,
+			"environment":       config.Environment,
+			"client_id":         config.ClientID,
+			"root_password_ttl": int(config.RootPasswordTTL.Seconds()),
```
austingebauer, line 190: Similar to what @fairclothjm mentioned in a different comment, I think we can maintain the `use_microsoft_graph_api` field in the read response. Its value can be hard-coded to `true`, since that's the only value that it can take after this point.

I could also see the argument for removing it completely, but let's see if that makes sense for what we've communicated in terms of deprecation. I'll check on this to be sure 👍

austingebauer (commented):
Looking good! Left a couple of suggestions. I'll take another pass on this shortly.

path_config.go
Zlaticanin: Thank you @austingebauer. I just pushed another PR where I removed the useMsGraphApi field completely and its usages. When you get a chance, could you please review it again? Thank you!!

austingebauer: Thanks and looks good! One last thing for this -- we should also remove the field from the path at [path_config.go#L74-L78](https://github.com/hashicorp/vault-plugin-secrets-azure/blob/removeaadgraphcode/path_config.go#L74-L78) since we're not parsing it anymore.

path_config.go
austingebauer: I'm starting to think it would be okay to remove the `use_microsoft_graph_api` field entirely for both writes and reads. Its purpose was to allow users to transition from AAD graph to MS graph during the deprecation phase. We've communicated the deprecation over the span of two releases (announcements went out in Vault 1.10 with the introduction of the field). It would serve no functional purpose after we've removed the AAD graph code. Let's see what others think.

fairclothjm: I'm good with removing `use_microsoft_graph_api` entirely for reads and writes.  👍 

My concern was that we would fail requests that provided the field. But I did some tests and it looks like we just ignore it. (This behavior in vault of ignoring unknown fields always catches me off guard.)

austingebauer: Thanks for the input, @fairclothjm! Sounds good to me. (+1 on the ignoring unknown fields point).

Issue #103, opened 2022-09-02
Permanently Delete SPs/Apps from AzureAD when Lease Expires


dpowley:
The Vault secrets engine does not permanently delete the service principals/apps from AzureAD when leases expire. Instead, the objects are placed in a "recycle bin," and they count toward the [limit of AzureAD objects in a tenant](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits#active-directory-limits) (that limit ranges from 50k to 500k objects). So after 50k-500k leases, Vault hits the limit of objects in an Azure AD and causes all create operations on the tenant to fail.

The Azure Vault secrets engine should at least provide the option to [permanently delete items](https://docs.microsoft.com/en-us/graph/api/directory-deleteditems-delete) upon expiration of leases.

austingebauer:
Closing as fixed by #104.

PR #104, opened 2022-09-07
Add option to permanently delete AzureAD objects

dpowley:
# Overview
This PR allows for permanent deletion of AzureAD apps and service principals that are created by the secrets engine. Traditionally, the Vault secrets engine does not permanently delete the service principals/apps from AzureAD when leases expire. Instead, the objects are placed in a "recycle bin," and they count toward the [limit of AzureAD objects in a tenant](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits#active-directory-limits) (that limit ranges from 50k to 500k objects). So after 50k-500k leases, Vault hits the limit of objects in an Azure AD and causes all create operations on the tenant to fail.

# Design of Change
How was this change implemented?
Added an optional field `permanently_delete` to the `role` path that defaults to `false` which invokes legacy behavior. If `permanently_delete` is set to `true`, the applications and service principals created by Vault will be permanently deleted when the corresponding leases expire.

[x] [Issue #103](https://github.com/hashicorp/vault-plugin-secrets-azure/issues/103)

# Contributor Checklist
[My Docs PR Link](link)
[x ] Backwards compatible

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -31,7 +31,7 @@ type AzureProvider interface {
 type ApplicationsClient interface {
 	GetApplication(ctx context.Context, applicationObjectID string) (ApplicationResult, error)
 	CreateApplication(ctx context.Context, displayName string) (ApplicationResult, error)
-	DeleteApplication(ctx context.Context, applicationObjectID string) error
+	DeleteApplication(ctx context.Context, applicationObjectID string, permanentlyDelete bool) error
 	ListApplications(ctx context.Context, filter string) ([]ApplicationResult, error)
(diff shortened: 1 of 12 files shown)


austingebauer (approved):
Thanks for this PR. I tested this out and successfully ran all tests in the repo with these changes. LGTM!

calvn:
@austingebauer can we add an entry to the CL on the repo so that we don't forget to note this change on the Vault side when we update the plugin version?

austingebauer:
@calvn - Will do! I'll open a PR here. I asked for a CL entry with the docs PR in Vault for this addition, so that won't be missed.

PR #110, opened 2022-11-07
Implement WAL rollback mechanism for Role Assignments

davidadeleon:
# Overview
Implement a rollback mechanism for Role Assignments to handle a case where on dynamic SP creation we may encounter an error and have to rollback any changes. Currently the AppID will be deleted which also removes the SP, but any successful role assignments will be orphaned. This incurs a hit against the Role Assignment limit. There is no change to the user experience.

# Design of Change
Create a separate WAL where the generated UUIDs of the role assignments can be stored, and referred to if needed for rollback. Added a TRACE level log for instances where a 204 response is returned from Azure. A 204 is returned when the Role Assignment ID has already been deleted or never existed.

# Test Output
```
❯ go test -v -run TestRoleAssignmentWALRollback
=== RUN   TestRoleAssignmentWALRollback/service_principals
=== PAUSE TestRoleAssignmentWALRollback/service_principals
=== CONT  TestRoleAssignmentWALRollback/service_principals
2022-11-07T13:37:24.357-0500 [DEBUG] rolling back role assignments for SP: ID=9cf990ff-39cb-4f20-b373-9afa49c9f752
2022-11-07T13:37:25.505-0500 [TRACE] role assignment already deleted or does not exist: err="error unassigning role: authorization.RoleAssignmentsClient#DeleteByID: Failure responding to request: StatusCode=204 -- Original Error: autorest/azure: error response cannot be parsed: \"\" error: EOF"
--- PASS: TestRoleAssignmentWALRollback (0.00s)
    --- PASS: TestRoleAssignmentWALRollback/service_principals (19.11s)
PASS
ok      github.com/hashicorp/vault-plugin-secrets-azure 19.352s
```


client.go
```diff
@@ -131,17 +131,12 @@ func (c *client) deleteServicePrincipal(ctx context.Context, spObjectID string,
...
-	for _, role := range roles {
-		assignmentID, err := uuid.GenerateUUID()
-		if err != nil {
-			return nil, err
-		}
-
+	for i, role := range roles {
 		resultRaw, err := retry(ctx, func() (interface{}, bool, error) {
-			ra, err := c.provider.CreateRoleAssignment(ctx, role.Scope, assignmentID,
+			ra, err := c.provider.CreateRoleAssignment(ctx, role.Scope, assignmentIDs[i],
```
jasonodonnell, line 139: Should we check if `[i]` exists? This could panic I think without a guard.

path_service_principal.go
```diff
@@ -133,11 +156,15 @@ func (b *azureSecretBackend) createSPSecret(ctx context.Context, s logical.Stora
...
 	}
 
-	// SP is fully created so delete the WAL
+	// SP is fully created so delete the WALs
 	if err := framework.DeleteWAL(ctx, s, walID); err != nil {
 		return nil, fmt.Errorf("error deleting WAL: %w", err)
 	}
 
+	if err := framework.DeleteWAL(ctx, s, rWALID); err != nil {
+		return nil, fmt.Errorf("error deleting WAL: %w", err)
```
jasonodonnell, line 165:
```suggestion
		return nil, fmt.Errorf("error deleting role assignment WAL: %w", err)
```

path_service_principal_test.go
```diff
@@ -545,6 +545,215 @@ func TestCredentialReadProviderError(t *testing.T) {
...
+
+		skipIfMissingEnvVars(t,
+			"AZURE_SUBSCRIPTION_ID",
+			"AZURE_CLIENT_ID",
+			"AZURE_CLIENT_SECRET",
+			"AZURE_TENANT_ID",
+		)
+
+		b := backend()
+		b.WALRollbackMinAge = 1 * time.Second
```
jasonodonnell, line 573: Think this is left over from trying to auto-trigger rollback code, could maybe be removed.

path_service_principal_test.go
```diff
@@ -545,6 +545,215 @@ func TestCredentialReadProviderError(t *testing.T) {
...
+		var assignmentIDs []string
+		for _, raID := range raIDs {
+			t := strings.Split(raID, "/")
+			tRa := t[len(t)-1]
+			assignmentIDs = append(assignmentIDs, strings.Replace(tRa, " ", "", -1))
+		}
+
+		// Remove one of the RA IDs to simulate a failure to assign a role
+		if err := client.unassignRoles(context.Background(), []string{raIDs[0]}); err != nil {
+			b.Logger().Error("error unassinging Role", "err", err)
```
jasonodonnell, line 679: Instead of using the logger here, we should put more context in t.Fatal (using t.Fatalf).

path_service_principal_test.go
```diff
@@ -545,6 +545,215 @@ func TestCredentialReadProviderError(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		rEntry, err := s.Get(context.Background(), fmt.Sprintf("%s/%s", "roles", roleName))
+		if err != nil {
+			t.Fatal(err)
```
jasonodonnell, line 685: In general the t.Fatal's in this test don't offer any context on what the test was doing before it failed, so it'll be harder to debug if they break.

wal.go
```diff
@@ -95,3 +100,59 @@ func (b *azureSecretBackend) rollbackRootWAL(ctx context.Context, req *logical.R
...
+		return err
+	}
+
+	b.Logger().Debug("rolling back role assignments for SP", "ID", entry.SpID)
+
+	// Assemble all App Role Assignment IDs
+	var roleAssignments []string
+	for i, assignmentID := range entry.AssignmentIDs {
+		roleAssignments = append(roleAssignments, fmt.Sprintf("%s/providers/Microsoft.Authorization/roleAssignments/%s",
+			entry.AzureRoles[i].Scope,
```
jasonodonnell, line 137: We might want a guard here to prevent panics.

path_service_principal.go
davidadeleon: Makes sense! Added this change.

path_service_principal_test.go
davidadeleon: Good catch. Cleaned up.

path_service_principal_test.go
davidadeleon: Used tFatalf instead to provide more context on a test failure.

wal.go
```diff
@@ -95,3 +100,62 @@ func (b *azureSecretBackend) rollbackRootWAL(ctx context.Context, req *logical.R
...
+	if err != nil {
+		return err
+	}
+
+	client, err := b.getClient(ctx, req.Storage)
+	if err != nil {
+		return err
+	}
+
+	b.Logger().Debug("rolling back role assignments for SP", "ID", entry.SpID)
```
jasonodonnell, line 131:
```suggestion
	b.Logger().Debug("rolling back role assignments for service principal", "ID", entry.SpID)
```

path_service_principal_test.go
```diff
@@ -545,6 +545,213 @@ func TestCredentialReadProviderError(t *testing.T) {
...
+		})
+		assertRespNoError(t, configResp, err)
+
+		roleName := "test_role_rawalrollback"
+
+		roleData := map[string]interface{}{
+			"azure_roles": fmt.Sprintf(`[
+			{
+				"role_name": "Storage Blob Data Owner",
+				"scope":  "/subscriptions/%s/resourceGroups/vault-azure-secrets-test1"
```
austingebauer, line 611: Might want to parameterize the resource group names via an env var as well?

client.go
```diff
@@ -131,17 +131,15 @@ func (c *client) deleteServicePrincipal(ctx context.Context, spObjectID string,
...
+func (c *client) assignRoles(ctx context.Context, spID string, roles []*AzureRole, assignmentIDs []string) ([]string, error) {
 	var ids []string
 
-	for _, role := range roles {
-		assignmentID, err := uuid.GenerateUUID()
-		if err != nil {
-			return nil, err
-		}
-
+	for i, role := range roles {
```
austingebauer, line 137: I'm thinking we should return an error if `len(roles) != len(assignmentIDs)` above this loop. That will ensure that we can safely use `assignmentIDs[i]` within the loop. The lengths are equal with our current usage of `assignRoles()`, but by adding the check, we avoid a panic if someone misuses this in the future.

path_service_principal_test.go
```diff
@@ -545,6 +545,213 @@ func TestCredentialReadProviderError(t *testing.T) {
 	}
 }
 
+// TestRoleAssignmentWALRollback tests rolling back any
+// role assignments that may have taken place prior to
+// a subsequent failure resulting in the need to rollback
+// an App or SP. This test requires valid, sufficiently-privileged
+// Azure credentials in env variables.
+func TestRoleAssignmentWALRollback(t *testing.T) {
```
austingebauer, line 553: Thanks for adding this!

wal.go
```diff
@@ -95,3 +100,67 @@ func (b *azureSecretBackend) rollbackRootWAL(ctx context.Context, req *logical.R
...
+	// Check any errors to filter out expected responses. Azure will return
+	// a 204 when trying to delete a role assignment that has already been
+	// deleted, or does not exist. We may hit this case during rollback.
+	if err := client.unassignRoles(ctx, roleAssignments); err != nil {
+		for _, e := range err.(*multierror.Error).Errors {
+			switch {
+			case strings.Contains(e.Error(), "StatusCode=204"):
+				b.Logger().Trace("role assignment already deleted or does not exist", "err", e.Error())
+			default:
+				b.Logger().Error("rollback error unassinging role", "err", e.Error())
```
austingebauer, line 158:
```suggestion
				b.Logger().Error("rollback error unassigning role", "err", e.Error())
```

austingebauer, line 158: Wondering if we should actually return the error in this case? I think we'll return `nil` and drop the WAL entry here when there was an error that was not a 204.

wal.go
```diff
@@ -95,3 +100,67 @@ func (b *azureSecretBackend) rollbackRootWAL(ctx context.Context, req *logical.R
...
+	client, err := b.getClient(ctx, req.Storage)
+	if err != nil {
+		return err
+	}
+
+	b.Logger().Debug("rolling back role assignments for service principal", "ID", entry.SpID)
+
+	// Return if there aren't any roles to unassign
+	if entry.AzureRoles == nil {
+		return fmt.Errorf("no azure roles associated with role")
```
austingebauer, line 135: If this entry will _always_ have `entry.AzureRoles == nil`, we might want to log and return nil instead? If we return an error here, the WAL may never get deleted.

austingebauer (commented):
Looking good! Left a few comments.

wal.go
davidadeleon: Logged the error, but returned nil here.

path_service_principal_test.go
```diff
@@ -608,12 +637,12 @@ func TestRoleAssignmentWALRollback(t *testing.T) {
...
 				"role_name": "Storage Blob Data Owner",
-				"scope":  "/subscriptions/%s/resourceGroups/vault-azure-secrets-test1"
+				"scope":  "/subscriptions/%s/resourceGroups/%s"
 			},
 			{
 				"role_name": "Reader",
-				"scope":  "/subscriptions/%s/resourceGroups/vault-azure-secrets-test2"
-			}]`, subscriptionID, subscriptionID),
+				"scope":  "/subscriptions/%s/resourceGroups/%s"
+			}]`, subscriptionID, subscriptionID, resourceGroup, resourceGroup),
```
jasonodonnell, line 645: Shouldn't this be `subscriptionID, resourceGroup, subscriptionID, resourceGroup`?

path_service_principal_test.go
```diff
@@ -129,37 +129,64 @@ func assertEmptyWAL(t *testing.T, b *azureSecretBackend, emp api.AzureProvider,
...
 
-		_, err = emp.GetApplication(context.Background(), app.AppObjID)
-		if err == nil {
-			t.Fatalf("expected error getting application")
+			_, err = emp.GetApplication(context.Background(), app.AppObjID)
+			if err == nil {
+				t.Fatalf("expected error getting application")
+			}
+		case walAppRoleAssignment:
+			fmt.Println(entry)
```
jasonodonnell, line 166: Left over debug line it looks like.

path_service_principal_test.go
davidadeleon: Yes it should be. This broke the test. Fixed this.

path_service_principal_test.go
davidadeleon: Cleaned this up.

path_service_principal_test.go
```diff
@@ -545,6 +571,215 @@ func TestCredentialReadProviderError(t *testing.T) {
...
+		// Serialize and deserialize the secret to remove typing, as will really happen.
+		fakeSaveLoad(credsResp.Secret)
+
+		// Revoke the Service Principal by sending back the secret we just received
+		req = &logical.Request{
+			Secret:  credsResp.Secret,
+			Storage: s,
+		}
+
+		b.spRevoke(context.Background(), req, nil)
```
austingebauer, line 774: Missing error check?

davidadeleon: Added an error check here.

wal.go
```diff
@@ -95,3 +100,68 @@ func (b *azureSecretBackend) rollbackRootWAL(ctx context.Context, req *logical.R
...
+		for _, e := range err.(*multierror.Error).Errors {
+			switch {
+			case strings.Contains(e.Error(), "StatusCode=204"):
+				b.Logger().Trace("role assignment already deleted or does not exist", "err", e.Error())
+			default:
+				return fmt.Errorf("rollback error unassinging role: %w", e)
+			}
+		}
+		if time.Now().After(entry.Expiration) {
+			return nil
```
austingebauer, line 163: I'm wondering if we should log something here to indicate that the WAL entry expired and there may be resources that aren't cleaned up 🤔 Looks like we don't above in `rollbackAppWAL`, but it might be useful to know when this happened.

austingebauer (approved):
LGTM!

client.go
```diff
@@ -299,9 +301,9 @@ func (b *azureSecretBackend) getClientSettings(ctx context.Context, config *azur
 
 // retry will repeatedly call f until one of:
 //
-//   * f returns true
-//   * the context is cancelled
-//   * 80 seconds elapses. Vault's default request timeout is 90s; we want to expire before then.
+//   - f returns true
```
vinay-gopalan, line 304: nit: these `fmt` changes should be fixed along with some dependency updates, might be worth updating with changes from `main` 😄

davidadeleon: Merged in changes from main.

PR #117, opened 2022-12-02
Fix: Use correct function signature when assigning roles

vinay-gopalan:
This PR adds a fix for an incorrect usage of the `assignRoles` function (not passing any `assignmentIDs`). The function signature was updated around the same time this particular usage was merged, and hence it led to a bug.

### DIFF
--- a/client.go
+++ b/client.go
@@ -14,10 +14,11 @@ import (
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/to"
-	multierror "github.com/hashicorp/go-multierror"
-	uuid "github.com/hashicorp/go-uuid"
-	"github.com/hashicorp/vault-plugin-secrets-azure/api"
+	"github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/go-uuid"
 	"github.com/hashicorp/vault/sdk/logical"
+
+	"github.com/hashicorp/vault-plugin-secrets-azure/api"
 )
 
 const (
@@ -139,6 +140,22 @@ func (c *client) deleteApp(ctx context.Context, appObjectID string, permanentlyD
 // deleteServicePrincipal deletes an Azure service principal.
 func (c *client) deleteServicePrincipal(ctx context.Context, spObjectID string, permanentlyDelete bool) error {
 	return c.provider.DeleteServicePrincipal(ctx, spObjectID, permanentlyDelete)
+}
+
+// generateAssignmentIDsFromRole pre-generates UUIDs to be
+// provided to assignRoles in case we need to rollback
+func (c *client) generateAssignmentIDsFromRole(role *roleEntry) ([]string, error) {
+	var assignmentIDs []string
+
+	for i := 0; i < len(role.AzureRoles); i++ {
+		assignmentID, err := uuid.GenerateUUID()
+		if err != nil {
+			return nil, err
+		}
+		assignmentIDs = append(assignmentIDs, assignmentID)
+	}
+
+	return assignmentIDs, nil
 }
 
 // assignRoles assigns Azure roles to a service principal.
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -141,6 +142,22 @@ func (c *client) deleteServicePrincipal(ctx context.Context, spObjectID string,
 	return c.provider.DeleteServicePrincipal(ctx, spObjectID, permanentlyDelete)
 }
 
+// generateAssignmentIDsFromRole pre-generates UUIDs to be
+// provided to assignRoles in case we need to rollback
+func (c *client) generateAssignmentIDsFromRole(role *roleEntry) ([]string, error) {
```
swenson, line 147: we're not actually using the roles, but only the length of the array. Maybe rework this to just be a simple function, `generateUUIDs`, that generates and returns a list of UUIDs of a certain length?

path_roles.go
```diff
@@ -345,13 +345,18 @@ func (b *azureSecretBackend) createPersistedApp(ctx context.Context, req *logica
 		return err
 	}
 
+	assignmentIDs, err := c.generateAssignmentIDsFromRole(role)
+	if err != nil {
+		return err
+	}
```
robmonte, line 351: Might be a good idea to wrap the error here.

Example:
```return fmt.Errorf("failed to generate assignmentIDs for PersistedApp roles: %w", err)```

Not sure if my example error text is accurate wording but hopefully it makes enough sense to convey my intention.

File: path_service_principal.go [hunk 3]
```diff
@@ -122,15 +121,9 @@ func (b *azureSecretBackend) createSPSecret(ctx context.Context, s logical.Stora
...
-
-	for i := 0; i < len(role.AzureRoles); i++ {
-		assignmentID, err := uuid.GenerateUUID()
-		if err != nil {
-			return nil, err
-		}
-		assignmentIDs = append(assignmentIDs, assignmentID)
+	assignmentIDs, err := c.generateAssignmentIDsFromRole(role)
+	if err != nil {
+		return nil, err
```
robmonte, line 126: Same thing about the error here for Service Principal.

robmonte (approved):
Just a small suggestion

client.go
vinay-gopalan: Good call! Updated in latest commit

path_roles.go
vinay-gopalan: Updated in latest commit

path_service_principal.go
vinay-gopalan: Updated in latest commit

swenson (approved):
LGTM

path_roles.go
```diff
@@ -345,13 +345,18 @@ func (b *azureSecretBackend) createPersistedApp(ctx context.Context, req *logica
 		return err
 	}
 
+	assignmentIDs, err := c.generateUUIDs(len(role.AzureRoles))
+	if err != nil {
+		return fmt.Errorf("error generating assginment IDs; err=%w", err)
```
austingebauer, line 350: Kind of a nit, but:
- Assignment is misspelled
- For consistency, I'd suggest to follow the general error format found throughout the secrets engine similar to `fmt.Errorf("error description: %w", err)`

path_roles.go
```diff
@@ -345,13 +345,18 @@ func (b *azureSecretBackend) createPersistedApp(ctx context.Context, req *logica
...
+	}
+
 	if role.ManagedApplicationObjectID != "" {
 		removeRolesAndGroupMembership(ctx, c, role)
 
 		spObjID := role.ServicePrincipalObjectID
 
 		// Assign Azure roles to the new SP
-		raIDs, err := c.assignRoles(ctx, spObjID, role.AzureRoles)
+		raIDs, err := c.assignRoles(ctx, spObjID, role.AzureRoles, assignmentIDs)
```
austingebauer, line 359: Related to how this integrates with the persisted app feature, I think this means that partial failure here wouldn't clean up role assignments? We have a mechanism for doing so introduced by https://github.com/hashicorp/vault-plugin-secrets-azure/pull/110. We should create a task to account for this in persisted app creation. We should be able to write this similar to what's done in https://github.com/hashicorp/vault-plugin-secrets-azure/blob/main/path_service_principal.go#L124-L144.

austingebauer (commented):
Sorry, a little late here 😄 A small nit and something we can likely create a different task for. Otherwise, LGTM!

Issue #92, opened 2022-03-06
Add support for Azure AD Role assignments to Vault Roles


ausfestivus:
Morning,

We would like to use the Azure Secrets Vault Plugin to generate an SP that can by used by the [Azure AD Terraform Provider](https://github.com/hashicorp/terraform-provider-azuread).

For this to work, we would need to be able to assign Azure AD Roles to the Vault Role definition. This doesn't appear to be currently supported. Note that we are talking about "Azure AD Roles", not "Azure Roles".

I asked about how to do this in the Azure AD Provider and they suggested [two alternate paths](https://github.com/hashicorp/terraform-provider-azuread/issues/738#issuecomment-1055832513).

I still think it would be advantageous to be able to create a Vault Role that can utilise the Azure AD Roles.

TimHodkin:
I have looked into this also and commented on another issue related to a similar issue:
https://github.com/hashicorp/vault-plugin-secrets-azure/issues/102#issuecomment-1433324023 

Something like this:
```
vault write azure/roles/my-role ttl=1h azure_roles=-<<EOF
    [
        {
            "role_name": "Global Reader",
            "scope": "/"
        }
    ]
EOF
```

PR #137, opened 2023-03-14
support Azure AD App roles

nbutton23:
# Overview

Support assigning [Azure AD App Roles](https://learn.microsoft.com/en-us/azure/active-directory/develop/howto-add-app-roles-in-azure-ad-apps) to the created Service Principal. This can be used to grant Service Principals access to the [Azure Graph API](https://learn.microsoft.com/en-us/graph/use-the-api). GraphAPI is the only supported way to perform actions in AAD

# Design of Change
Adds a new field `app_roles` that accepts an array of apps and the roles to be assigned.  Using the [Graph API appRoleAssignment](https://learn.microsoft.com/en-us/graph/api/user-post-approleassignments?view=graph-rest-1.0&tabs=http) the roles are added.  

Roles are expected to be provided as a JSON formatted string in the format 

```
"app_roles": “[
        {   
                \"app_id\": \"<uuid>\",
                \"roles\": [
                        {
                                \"role_name\": \"Directory.Read.All\"
                        }
                ]
        }
]”
```

I have been running a version of this plugin on our vault instance for a few months with no issues. 

[ ] [Issue #92](https://github.com/hashicorp/vault-plugin-secrets-azure/issues/92)

# Contributor Checklist
[My Docs PR Link](https://github.com/hashicorp/vault/pull/19537)
[x] Backwards compatible


api/application_msgraph.go
```diff
@@ -323,9 +323,11 @@ func (c AppClient) removePasswordResponder(resp *http.Response) (autorest.Respon
 
 func (c AppClient) createApplicationPreparer(ctx context.Context, displayName string) (*http.Request, error) {
 	parameters := struct {
-		DisplayName *string `json:"displayName"`
+		DisplayName    *string `json:"displayName"`
+		SignInAudience *string `json:"signInAudience"`
 	}{
-		DisplayName: to.StringPtr(displayName),
+		DisplayName:    to.StringPtr(displayName),
+		SignInAudience: to.StringPtr("AzureADMyOrg"),
```
nbutton23, line 330: This is needed to fix an issue with `persist_app` since the default value will only allow 2 active tokens at a time.  

If needed I can pull this into its own PR.

PR #146, opened 2023-05-02
use common bulk dependency upgrades workflow

fairclothjm:
- use common bulk dependency upgrades workflow
- remove dependabot


maxcoulombe:
Action linting is unhappy but it doesn't seem to be related to your changes?

PR #150, opened 2023-07-04
fix: only run periodic func on active instance in primary cluster or local mount

austingebauer:
## Overview

This PR adds a condition to the periodic func that ensures it only runs on the active instance in a primary cluster or on local mounts. This fixes an issue where root credential rotation was taking place on the active/secondary _and_ active/primary. This caused intermittent 401s due to the rotations racing and invalidating the credentials of each other's cached clients.

## Testing

I was able to reproduce this by getting the timing right on the active/secondary's periodicFunc such that it would perform the rotation. This also resulted in a "cannot write to readonly storage" error within the periodic func (see image below).

Screenshot 2023-07-03 at 5 36 12 PM

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/hashicorp/vault-plugin-secrets-azure/api"
 	"github.com/hashicorp/vault/sdk/framework"
+	"github.com/hashicorp/vault/sdk/helper/consts"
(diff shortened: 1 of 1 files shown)


elliesterner (approved):

fairclothjm (approved):
LGTM!

PR #161, opened 2023-09-06
Update to v0.68.0 and the finalized/non preview management API, to be compatible with other consumers of the Azure SDK

sgmiller:
# Overview
When updating other plugins that use the Azure SDK, we hit an
error as the Azure secrets engine imports a deprecated preview API.
This PR updates to the compatible Azure SDK and updates to the
nearest final API

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -7,7 +7,7 @@ import (
 	"context"
 	"time"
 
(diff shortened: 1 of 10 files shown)


fairclothjm (approved):

austingebauer (approved):
Thanks!

Issue #166, opened 2023-10-24
Migrate away from Azure/go-autorest


F21:
The go-autorest has been deprecated and unsupported as of March 31 2023.
See: 
- https://github.com/Azure/go-autorest
- https://azure.microsoft.com/en-us/updates/support-for-azure-sdk-libraries-that-do-not-conform-to-our-current-azure-sdk-guidelines-will-be-retired-as-of-31-march-2023/

The plugin should migrate to using the [azidentity](https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/azidentity) package.

PR #169, opened 2023-10-30
Replace `go-autorest` MS Graph client with `msgraph-sdk-go`

vinay-gopalan:
# Overview
This PR replaces the deprecated [go-autorest](https://github.com/Azure/go-autorest) module with [msgraph-sdk-go](https://github.com/microsoftgraph/msgraph-sdk-go). 

[ ] #166 
[ ] [PR #1234](https://github.com/hashicorp/vault/pr/1234)

# Contributor Checklist
[x] Backwards compatible

```
$ make test
--- PASS: TestPeriodicFuncNilConfig (0.00s)
=== RUN   TestRetry
=== PAUSE TestRetry
=== RUN   TestConfig
=== RUN   TestConfig/root_password_ttl_defaults_to_6_months
=== RUN   TestConfig/root_password_ttl_set_if_provided
=== RUN   TestConfig/environment_set_if_provided
--- PASS: TestConfig (0.00s)
    --- PASS: TestConfig/root_password_ttl_defaults_to_6_months (0.00s)
    --- PASS: TestConfig/root_password_ttl_set_if_provided (0.00s)
    --- PASS: TestConfig/environment_set_if_provided (0.00s)
=== RUN   TestConfigEnvironmentClouds
--- PASS: TestConfigEnvironmentClouds (0.00s)
=== RUN   TestConfigDelete
--- PASS: TestConfigDelete (0.00s)
=== RUN   TestRoleCreate
=== RUN   TestRoleCreate/SP_role
=== RUN   TestRoleCreate/SP_persistent_role
=== RUN   TestRoleCreate/Static_SP_role
=== RUN   TestRoleCreate/Optional_role_TTLs
=== RUN   TestRoleCreate/Role_TTL_Checks
=== RUN   TestRoleCreate/Role_name_lookup
=== RUN   TestRoleCreate/Group_name_lookup
=== RUN   TestRoleCreate/Duplicate_role_name_and_scope
=== RUN   TestRoleCreate/Duplicate_role_name,_different_scope
=== RUN   TestRoleCreate/Duplicate_group_object_ID
=== RUN   TestRoleCreate/Role_name_lookup_(multiple_match)
=== RUN   TestRoleCreate/Group_name_lookup_(multiple_match)
--- PASS: TestRoleCreate (0.00s)
    --- PASS: TestRoleCreate/SP_role (0.00s)
    --- PASS: TestRoleCreate/SP_persistent_role (0.00s)
    --- PASS: TestRoleCreate/Static_SP_role (0.00s)
    --- PASS: TestRoleCreate/Optional_role_TTLs (0.00s)
    --- PASS: TestRoleCreate/Role_TTL_Checks (0.00s)
    --- PASS: TestRoleCreate/Role_name_lookup (0.00s)
    --- PASS: TestRoleCreate/Group_name_lookup (0.00s)
    --- PASS: TestRoleCreate/Duplicate_role_name_and_scope (0.00s)
    --- PASS: TestRoleCreate/Duplicate_role_name,_different_scope (0.00s)
    --- PASS: TestRoleCreate/Duplicate_group_object_ID (0.00s)
    --- PASS: TestRoleCreate/Role_name_lookup_(multiple_match) (0.00s)
    --- PASS: TestRoleCreate/Group_name_lookup_(multiple_match) (0.00s)
=== RUN   TestRoleCreateBad
--- PASS: TestRoleCreateBad (0.00s)
=== RUN   TestRoleUpdateError
--- PASS: TestRoleUpdateError (0.00s)
=== RUN   TestRoleList
--- PASS: TestRoleList (0.00s)
=== RUN   TestRoleDelete
--- PASS: TestRoleDelete (0.00s)
=== RUN   TestRotateRootSuccess
    path_rotate_root_test.go:19: Missing env variable: [AZURE_CLIENT_ID] - skipping test
--- SKIP: TestRotateRootSuccess (0.00s)
=== RUN   TestRotateRootPeriodicFunctionBeforeMinute
    path_rotate_root_test.go:98: Missing env variable: [AZURE_CLIENT_ID] - skipping test
--- SKIP: TestRotateRootPeriodicFunctionBeforeMinute (0.00s)
=== RUN   TestSP_WAL_Cleanup
=== RUN   TestSP_WAL_Cleanup/Role_assign_fail
--- PASS: TestSP_WAL_Cleanup (5.00s)
    --- PASS: TestSP_WAL_Cleanup/Role_assign_fail (5.00s)
=== RUN   TestSPRead
=== RUN   TestSPRead/Basic_Role
=== RUN   TestSPRead/Basic_Group
=== RUN   TestSPRead/TTLs
--- PASS: TestSPRead (0.00s)
    --- PASS: TestSPRead/Basic_Role (0.00s)
    --- PASS: TestSPRead/Basic_Group (0.00s)
    --- PASS: TestSPRead/TTLs (0.00s)
=== RUN   TestStaticSPRead
=== RUN   TestStaticSPRead/Basic
=== RUN   TestStaticSPRead/TTLs
--- PASS: TestStaticSPRead (0.00s)
    --- PASS: TestStaticSPRead/Basic (0.00s)
    --- PASS: TestStaticSPRead/TTLs (0.00s)
=== RUN   TestPersistentAppSPRead
=== RUN   TestPersistentAppSPRead/Basic
=== RUN   TestPersistentAppSPRead/TTLs
--- PASS: TestPersistentAppSPRead (0.00s)
    --- PASS: TestPersistentAppSPRead/Basic (0.00s)
    --- PASS: TestPersistentAppSPRead/TTLs (0.00s)
=== RUN   TestSPRevoke
=== RUN   TestSPRevoke/roles
=== RUN   TestSPRevoke/permanently_delete_roles
=== RUN   TestSPRevoke/groups
--- PASS: TestSPRevoke (0.00s)
    --- PASS: TestSPRevoke/roles (0.00s)
    --- PASS: TestSPRevoke/permanently_delete_roles (0.00s)
    --- PASS: TestSPRevoke/groups (0.00s)
=== RUN   TestStaticSPRevoke
--- PASS: TestStaticSPRevoke (0.00s)
=== RUN   TestSPReadMissingRole
--- PASS: TestSPReadMissingRole (0.00s)
=== RUN   TestCredentialReadProviderError
--- PASS: TestCredentialReadProviderError (0.00s)
--- SKIP: TestRoleAssignmentWALRollback (0.00s)
=== RUN   TestCredentialInteg_msgraph
--- SKIP: TestCredentialInteg_msgraph (0.00s)
=== CONT  TestRetry
=== RUN   TestRetry/First_try_success
=== RUN   TestRetry/Three_retries
=== PAUSE TestRetry/Three_retries
=== RUN   TestRetry/Error_on_attempt
=== PAUSE TestRetry/Error_on_attempt
=== RUN   TestRetry/Timeout
=== PAUSE TestRetry/Timeout
=== RUN   TestRetry/Cancellation
=== PAUSE TestRetry/Cancellation
=== CONT  TestRetry/Three_retries
=== CONT  TestRetry/Timeout
=== CONT  TestRetry/Cancellation
=== CONT  TestRetry/Error_on_attempt
--- PASS: TestRetry (0.00s)
    --- PASS: TestRetry/First_try_success (0.00s)
    --- PASS: TestRetry/Error_on_attempt (0.00s)
    --- PASS: TestRetry/Cancellation (1.00s)
    --- PASS: TestRetry/Timeout (10.00s)
    --- PASS: TestRetry/Three_retries (12.00s)
PASS
ok      github.com/hashicorp/vault-plugin-secrets-azure 18.475s
?       github.com/hashicorp/vault-plugin-secrets-azure/api     [no test files]
?       github.com/hashicorp/vault-plugin-secrets-azure/cmd/vault-plugin-secrets-azure  [no test files]
?       github.com/hashicorp/vault-plugin-secrets-azure/mocks   [no test files]

```


mock_provider_test.go
```diff
@@ -1,260 +0,0 @@
-// Code generated by MockGen. DO NOT EDIT.
```
vinay-gopalan, line 1: Did not see this being used anywhere. Plus we have another `provider_mock_test` which seems to be achieving the same thing. Opted to delete this and update the latter with new MSGraph SDK mock tests

client.go
```diff
@@ -208,11 +209,12 @@ func (c *client) unassignRoles(ctx context.Context, roleIDs []string) error {
 
 	for _, id := range roleIDs {
 		if _, err := c.provider.DeleteRoleAssignmentByID(ctx, id); err != nil {
-			detailedErr := new(autorest.DetailedError)
 			// If a role was deleted manually then Azure returns a error and status 204
-			if errors.As(err, detailedErr) && (detailedErr.StatusCode == http.StatusNoContent || detailedErr.StatusCode == http.StatusNotFound) {
-				continue
-			}
+			// @TODO figure out how to get the status code from MSGraph SDK errors; removing autorest for now
```
vinay-gopalan, line 213: Opted to remove autorest for the time being, but not sure how to best replicate this with the tools at hand. I think the error returned from MSGraph should contain a status code to perform the logic below, but need to double check this

fairclothjm: I think you can do something similar to what we did in TFVP: https://github.com/hashicorp/terraform-provider-vault/pull/2050/files#diff-de5e35945f06f622d0b58ae2f696ead44ed39d0acb73fc826309a03ce4e00b83R252-R262

Which used https://github.com/Azure/azure-sdk-for-go/blob/main/documentation/MIGRATION_GUIDE.md#error-handling as the guide

fairclothjm: Not sure if that will work with MSGraph?

mocks/Applicationable.go
fairclothjm, line 1: Should we add a makefile target, readme section or something similar to describe how these mocks were generated and what commands can be used to update them if needed?

client.go
vinay-gopalan: Nice, it looks to be in the `azcore` library, so I think it should work (🤞🏼) Thanks for the reference 🙏🏼

provider_mock_test.go
```diff
@@ -102,40 +116,48 @@ func (m *mockProvider) CreateServicePrincipal(_ context.Context, _ string, _ tim
 	return id, pass, nil
 }
 
-func (m *mockProvider) CreateApplication(_ context.Context, _ string) (api.ApplicationResult, error) {
+func (m *mockProvider) CreateApplication(_ context.Context, _ string) (models.Applicationable, error) {
+	if m.ctxTimeout != 0 {
```
vinay-gopalan, line 120: I added a time.Sleep to simulate the context deadline error expected by [`TestSP_WAL_Cleanup`  ](https://github.com/hashicorp/vault-plugin-secrets-azure/pull/169/files#diff-77464a25e41d440bcffea845c2842cd0c3d57deafa53c2391826d780b7bde89aR96)

mocks/Applicationable.go
vinay-gopalan: I think that is a good idea! I can add a note in the README under the Testing section describing how the mocks were generated using `mockery`

provider_mock_test.go
raymonstah: If we keep this approach, can we make `ctxTimeout` a `time.Duration` and remove the hardcoded `time.Second` below?

raymonstah (commented):
Great work! Had some concerns about returning the msgraph package interface and exposing it in our (presumably public) interface.

api/applications.go
```diff
@@ -0,0 +1,168 @@
...
+	if len(apps) > 1 {
+		return nil, fmt.Errorf("multiple applications found - double check your client_id")
+	}
+
+	return apps[0], nil
+}
+
+func (c *AppClient) ListApplications(ctx context.Context, filter string) ([]models.Applicationable, error) {
+	headers := abstractions.NewRequestHeaders()
+	headers.Add("ConsistencyLevel", "eventual")
```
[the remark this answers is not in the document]
fairclothjm: Doesn't look like we actually need it? Since we are only ever using the "eq" filter? https://learn.microsoft.com/en-us/graph/aad-advanced-queries?tabs=go

provider_mock_test.go
vinay-gopalan: Sure thing, I can update that 👍🏼

api/applications.go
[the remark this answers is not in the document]
vinay-gopalan: Ah interesting, thanks for sending that link for reference! I think we might be okay to not use the header, I was mainly going off of the official [MSGraph API documentation](https://learn.microsoft.com/en-us/graph/api/application-list?view=graph-rest-1.0&tabs=go#example-3-use-filter-and-top-to-get-one-application-with-a-display-name-that-starts-with-a-including-a-count-of-returned-objects), where they seem to be setting the header for the List Applications call even though they are using a filter 🤔 

I'll test without the header and if it works I'll remove it 👍🏼

api/applications.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/google/uuid"
+	"github.com/hashicorp/go-multierror"
+	abstractions "github.com/microsoft/kiota-abstractions-go"
+	msgraphsdkgo "github.com/microsoftgraph/msgraph-sdk-go"
+	auth "github.com/microsoftgraph/msgraph-sdk-go-core/authentication"
+	"github.com/microsoftgraph/msgraph-sdk-go/applications"
+	"github.com/microsoftgraph/msgraph-sdk-go/models"
+)
+
+type ApplicationsClient interface {
```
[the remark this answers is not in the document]
vinay-gopalan: Great thoughts!

I'm curious about the idea of defining our own types — I had a similar thought initially, but was having a hard time figuring out how to get our constructed types to match up with the complicated interfaces (such as `models.Applicationable`) that Microsoft API calls return. Did you have ideas on how we might define the types? Would definitely make our lives a lot easier if we do not have to keep track of generated mock types

api/applications.go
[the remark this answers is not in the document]
vinay-gopalan: Ah, now that I read the docs more closely, it _might_ be that we only need that header if we use orderby. Y'all are right, looks like we don't need the header! Thanks for the catch 🙏🏼

api/applications.go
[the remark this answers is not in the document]
raymonstah: The types defined could just be regular Go structs:
```
type Application struct {
	ID          string
	AppID       string
	Description string
	DisplayName string
}

func (c *AppClient) GetApplication(ctx context.Context, clientID string) (Application, error) {
	filter := fmt.Sprintf("appId eq '%s'", clientID)
	req := applications.ApplicationsRequestBuilderGetRequestConfiguration{
		QueryParameters: &applications.ApplicationsRequestBuilderGetQueryParameters{
			Filter: &filter,
		},
	}

	resp, err := c.client.Applications().Get(ctx, &req)
	if err != nil {
		return Application{}, err
	}

	apps := resp.GetValue()
	if len(apps) == 0 {
		return Application{}, fmt.Errorf("no application found")
	}
	if len(apps) > 1 {
		return Application{}, fmt.Errorf("multiple applications found - double check your client_id")
	}
	app := apps[0]
	application := Application{
		ID:          *app.GetId(),
		AppID:       *app.GetAppId(),
		Description: *app.GetDescription(),
		DisplayName: *app.GetDisplayName(),
	}

	return application, nil
}
```

api/applications.go
```diff
@@ -0,0 +1,168 @@
...
+var _ ApplicationsClient = (*AppClient)(nil)
+var _ GroupsClient = (*AppClient)(nil)
+var _ ServicePrincipalClient = (*AppClient)(nil)
+
+type AppClient struct {
+	client *msgraphsdkgo.GraphServiceClient
+}
+
+// Reference: https://docs.microsoft.com/en-us/graph/deployments#microsoft-graph-and-graph-explorer-service-root-endpoints
+func GetGraphURI(env string) (string, error) {
```
austingebauer, line 39: I think this can be removed? Looks replaced by `graphURIFromName`.

vinay-gopalan: Yup!

api/applications.go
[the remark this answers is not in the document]
vinay-gopalan: Ah I see what you mean! I think this might work, I might have overcomplicated things in my head initially. Going to give this refactoring a try, might not end up needing the mocks this way 🙏🏼

provider.go
```diff
@@ -5,110 +5,189 @@ package azuresecrets
...
 	groupsClient api.GroupsClient
-	raClient     *authorization.RoleAssignmentsClient
-	rdClient     *authorization.RoleDefinitionsClient
+	raClient     *armauthorization.RoleAssignmentsClient
+	rdClient     *armauthorization.RoleDefinitionsClient
 }
 
 // newAzureProvider creates an azureProvider, backed by Azure client objects for underlying services.
-func newAzureProvider(settings *clientSettings, passwords api.Passwords) (api.AzureProvider, error) {
+func newAzureProvider(settings *clientSettings, passwords api.Passwords) (AzureProvider, error) {
```
austingebauer, line 66: This is unrelated to your PR (sorry!) but probably a good time to figure out what's going on. It appears that `passwords` is unused here. Looking closer, it appears that our password `Generate()` func in [api/passwords.go](https://github.com/hashicorp/vault-plugin-secrets-azure/blob/VAULT-11806/msgraph-sdk-upgrade/api/passwords.go#L26) is also unused. It doesn't looks like Azure lets you provide a password but instead returns one to the client. This leads me to wonder if/how [password_policy](https://github.com/hashicorp/vault-plugin-secrets-azure/blob/main/path_config.go#L76) works? 🤔

fairclothjm: It doesn't appear as though it does work. I wonder if we need to use [SetSecretText](https://github.com/microsoftgraph/msgraph-sdk-go/blob/9f7263506f34618b4206543ff2115e96b1233d72/models/password_credential.go#L328) in AddApplicationPassword()?

raymonstah: Based on their API docs:
1. https://learn.microsoft.com/en-us/graph/api/application-addpassword?view=graph-rest-1.0&tabs=http
2. https://learn.microsoft.com/en-us/graph/api/serviceprincipal-addpassword?view=graph-rest-1.0&tabs=http

It doesn't seem like there's a way to set the password.

Maybe the fields were just added in case their API adds support for a user defined password in the future.

api/groups.go
```diff
@@ -5,16 +5,56 @@ package api
...
+	GetGroup(ctx context.Context, objectID string) (result models.Groupable, err error)
+	ListGroups(ctx context.Context, filter string) (result []models.Groupable, err error)
 }
 
 type Group struct {
 	ID          string
 	DisplayName string
 }
+
+func (c *AppClient) AddGroupMember(ctx context.Context, groupObjectID string, memberObjectID string) error {
```
austingebauer, line 26: Should we be using `memberObjectID` in the request somewhere?

api/applications.go
```diff
@@ -0,0 +1,168 @@
...
+	ListApplications(ctx context.Context, filter string) ([]models.Applicationable, error)
+	AddApplicationPassword(ctx context.Context, applicationObjectID string, displayName string, endDateTime time.Time) (models.PasswordCredentialable, error)
+	RemoveApplicationPassword(ctx context.Context, applicationObjectID string, keyID *uuid.UUID) error
+}
+
+var _ ApplicationsClient = (*AppClient)(nil)
+var _ GroupsClient = (*AppClient)(nil)
+var _ ServicePrincipalClient = (*AppClient)(nil)
+
+type AppClient struct {
```
austingebauer, line 34: I wonder if `MSGraphClient` is a better name for this? It looks like it's the receiver for methods throughout apps, groups, SPs.

api/service_principals.go
```diff
@@ -14,7 +20,67 @@ type ServicePrincipalClient interface {
...
+
+	spList, err := c.client.ServicePrincipals().Get(ctx, configuration)
+	if err != nil {
+		return nil, err
+	}
+	return spList.GetValue(), nil
+}
+
+func (c *AppClient) GetServicePrincipalByID(ctx context.Context, spObjectID string) (models.ServicePrincipalable, error) {
+	return c.client.ServicePrincipals().ByServicePrincipalId(spObjectID).Get(ctx, nil)
```
austingebauer, line 85: It looks like these are only used in test code. I'd suggest to keep them in the test code to keep the interface minimal and have a clean separation between the test code and the plugin code.

austingebauer (commented):
Looking good so far! Haven't made it all the way through yet. Will circle back tomorrow.

api/applications.go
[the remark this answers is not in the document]
vinay-gopalan: @raymonstah thanks for the extra set of eyes! The refactoring definitely made things simpler and removed the need for mocks altogether 🎉

mocks/Applicationable.go
vinay-gopalan: Using internal types allowed us to avoid using mocks! I was initially trying a more complicated approach, but for long term, I realized it's probably better to move forward without generating mocks for the complex interfaces 😅

api/groups.go
vinay-gopalan: Yup! Forgot to encode it as part of the `oData`. Fixed now

provider_mock_test.go
vinay-gopalan: Updated!

provider.go
```diff
@@ -5,110 +5,179 @@ package azuresecrets
...
-	"github.com/Azure/go-autorest/autorest/azure/auth"
-	"github.com/hashicorp/vault-plugin-secrets-azure/api"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/hashicorp/go-cleanhttp"
+	"github.com/hashicorp/vault/sdk/logical"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization"
```
austingebauer, line 19: I believe we're going to need to upgrade this to `github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization/v2` per guidance in https://github.com/Azure/azure-sdk-for-go/issues/18674. The reason is that when running the acceptance tests you'll receive an error like `"UnsupportedApiVersionForRoleDefinitionHasDataActions"` when assigning roles with dataActions/notDataActions. This is a behavior we'll want to maintain (had a customer issue when we accidentally removed it in the past, hence the test coverage for it 😅).

It appears there is some breakage (mostly struct naming?) in the v2 package that'll need some adjustments. Lmk if you need any help or if I can explain better!

provider.go
austingebauer: Hmm, ya it doesn't look like you can set a password even with that `SetSecretText()` method. I think it's safe to say password policies don't work. I'm thinking we'll want to deprecate the parameter, remove the code, and announce this in documentation / upgrade guides at minimum. Can be done in a separate task if that's preferred.

api/service_principals.go
vinay-gopalan: I was having trouble keeping the interface for service principals to the minimal (currently at two methods) when I attempted to move these methods to the test file and define a test interface. I think the moving of these to a test file would require 1 additional method at the least to the interface — something like `GetSPClient` that can be called from the test file to convert a service principal client to a `testSPClient` interface that supports the additional two methods.

Maybe I was looking at things in the wrong way, wrangling all of the different interfaces might have led me down a different path. Would be open to your thoughts on what you think the test file interfaces could look like while maintaining the minimal interface! 🙏🏼  Since these 2 functions are only being used in acceptance tests, I mainly need access to the `MSGraphClient.client *msgraphsdgo.GraphServiceClient`, which seems hard to do without adding an interface method.

I did clean up a little bit of the implementation by adding internal types and not relying on `models.ServicePrincipalable`

provider.go
vinay-gopalan: Updated!

client.go
```diff
@@ -185,8 +188,7 @@ func (c *client) assignRoles(ctx context.Context, spID string, roles []*AzureRol
 			if err != nil && strings.Contains(err.Error(), "PrincipalNotFound") {
 				return nil, false, nil
 			}
-
-			return to.String(ra.ID), true, err
+			return *ra.ID, true, err
```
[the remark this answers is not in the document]
vinay-gopalan: Added an empty response check case [here](https://github.com/hashicorp/vault-plugin-secrets-azure/pull/169/files#diff-4b667feae66c9d46b21b9ecc19e8958cf4472d162ce0a47ac3e8386af8bbd8cfR193). Let me know if this resolves the panic you were seeing! 

Also, I removed `to.String` from a bunch of places, wondering if this behavior needs to be replicated elsewhere?

vinay-gopalan:
Not sure why the `go mod tidy` checks are failing. It started when I switched to using internal types for MSGraph SDK responses and removed the mocks. Local runs of `go mod tidy` don't lead to any changes 🤔

provider.go
vinay-gopalan: Thanks for the context folks! Will add a ticket to track this in our backlog, and going to opt to tackle this in a follow-up PR 👍🏼

provider.go
```diff
@@ -137,47 +205,37 @@ func (p *provider) DeleteServicePrincipal(ctx context.Context, spObjectID string
...
 
 // DeleteRoleAssignmentByID deletes a role assignment.
-func (p *provider) DeleteRoleAssignmentByID(ctx context.Context, roleAssignmentID string) (result authorization.RoleAssignment, err error) {
-	return p.raClient.DeleteByID(ctx, roleAssignmentID)
-}
-
-// ListRoleAssignments lists all role assignments.
-// There is no need for paging; the caller only cares about the first match and whether
-// there are 0, 1 or >1 items. Unpacking here is a simpler interface.
-func (p *provider) ListRoleAssignments(ctx context.Context, filter string) ([]authorization.RoleAssignment, error) {
```
vinay-gopalan, line 173: Did not see thing being used anywhere, opted to remove

client.go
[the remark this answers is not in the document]
austingebauer: Taking a look around and will let you know what I find 👀

provider.go
austingebauer: Tracked this back a bit and looks like it was used for testing before. Good with removing it 👍

api/service_principals.go
```diff
@@ -15,6 +21,87 @@ type ServicePrincipalClient interface {
...
+	filter := fmt.Sprintf("appId eq '%s'", spObjectID)
+	requestParameters := &serviceprincipals.ServicePrincipalsRequestBuilderGetQueryParameters{
+		Filter: &filter,
+	}
+
+	headers := abstractions.NewRequestHeaders()
+	headers.Add("ConsistencyLevel", "eventual")
+
+	configuration := &serviceprincipals.ServicePrincipalsRequestBuilderGetRequestConfiguration{
+		Headers:         headers,
```
fairclothjm, line 78: I believe we can remove this as well since we are only ever using the `eq` filter.

api/applications.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	var apps []Application
+	for _, app := range resp.GetValue() {
+		apps = append(apps, Application{
+			ID:                  *app.GetId(),
+			AppID:               *app.GetAppId(),
+			AppObjectID:         *app.GetId(),
+			Description:         *app.GetDescription(),
+			DisplayName:         *app.GetDisplayName(),
```
fairclothjm, line 136: If any of these are nil we will get a panic trying to dereference them. I got a panic running `make testacc` on `*app.GetDescription()`. Maybe we should add thin wrappers functions for these that will return zero values if the dereference is nil? Also, I don't see any reason to keep store `ID`, `Description`, and `DisplayName`. From what I can tell we only make use of `AppID` and `AppObjectID`. So I think removing the unused fields could make this work simpler.

path_service_principal_test.go
```diff
@@ -979,23 +980,23 @@ func TestCredentialInteg_msgraph(t *testing.T) {
...
 		// info from Azure and verify it matches the Reader role.
 		raIDs := credsResp.Secret.InternalData["role_assignment_ids"].([]string)
 		equal(t, 2, len(raIDs))
 
-		ra, err := provider.raClient.GetByID(context.Background(), raIDs[0])
+		ra, err := provider.raClient.GetByID(context.Background(), raIDs[0], nil)
 		assertErrorIsNil(t, err)
 
-		roleDefs, err := provider.ListRoleDefinitions(context.Background(), fmt.Sprintf("subscriptions/%s", subscriptionID), "")
+		roleDefs, err := provider.ListRoleDefinitions(nil, fmt.Sprintf("subscriptions/%s", subscriptionID), "")
```
fairclothjm, line 993: We need to pass a context here, otherwise we get `unexpected error: net/http: nil Context`

api/applications.go
vinay-gopalan: Ah that is a good point and wrappers are a great idea. Added thin wrapper helpers to return zero values in case the responses from MS Graph API are `nil` values. Also cut down the structs `Application` and `PasswordCredential` to only required fields. Thanks for the great points!

api/applications.go
```diff
@@ -210,12 +178,38 @@ func (c *MSGraphClient) RemoveApplicationPassword(ctx context.Context, applicati
...
+		return Application{
+			AppID:               "",
+			AppObjectID:         "",
+			PasswordCredentials: []PasswordCredential{},
+		}
+
+	}
+	return Application{
+		AppID:               *app.GetAppId(),
+		AppObjectID:         *app.GetId(),
```
fairclothjm, line 197: In theory it is still possible to get a panic if GetId or GetAppId return nil. In my case, the panic I got was when `app` as non-nil but GetDescription returned nil. So we should probably make sure these function calls are non-nil before we dereference them just in case. :)

path_service_principal_test.go
```diff
@@ -979,23 +980,23 @@ func TestCredentialInteg_msgraph(t *testing.T) {
...
 		assertErrorIsNil(t, err)
 
-		defID := *ra.RoleAssignmentPropertiesWithScope.RoleDefinitionID
+		defID := *ra.Properties.RoleDefinitionID
 		found := false
 		for _, def := range roleDefs {
-			if *def.ID == defID && *def.RoleName == "Storage Blob Data Owner" {
+			if *def.ID == defID && *def.Name == "Storage Blob Data Owner" {
```
fairclothjm, line 999: This should be
```suggestion
			if *def.ID == defID && *def.Properties.RoleName == "Storage Blob Data Owner" {
```

path_service_principal_test.go
```diff
@@ -1019,7 +1020,7 @@ func TestCredentialInteg_msgraph(t *testing.T) {
 		// Verify that SP get is an error after delete. Expected there
 		// to be a delay and that this step would take some time/retries,
 		// but that seems not to be the case.
-		assertServicePrincipalDoesNotExist(t, provider.spClient, spObjID)
+		assertServicePrincipalExists(t, provider.spClient, spObjID, false)
```
fairclothjm, line 1023: this is failing with
```
    path_service_principal_test.go:1023:
        unexpected error: Resource '1252c6cf-ca96-4de1-ad73-c7b60cea200c' does not exist or one of its queried reference-property objects are not present.
--- FAIL: TestCredentialInteg_msgraph (0.00s)
    --- FAIL: TestCredentialInteg_msgraph/service_principals (27.03s)
    
```

Should we be asserting that it Does Not Exist?

fairclothjm: Oh, I see we are passing false for the last parameter. I missed that. Should we separate them into separate functions like they were before? Either way, I think we will want to not fail when we get an error in this helper if we are expecting the resource to not exist.

vinay-gopalan: Fixed in latest commit. I tested things locally and the test passes now after ensuring the correct behavior on resource non-existence. Thanks!

path_service_principal_test.go
vinay-gopalan: Nice catch, updated

api/applications.go
vinay-gopalan: Ah interesting, good to know. I added some extra guardrails on the function calls themselves to prevent any nil pointer dereferencing :)

backend_test.go
```diff
@@ -44,7 +45,8 @@ func getTestBackendMocked(t *testing.T, initConfig bool) (*azureSecretBackend, l
 
 	b.settings = new(clientSettings)
 	mockProvider := newMockProvider()
-	b.getProvider = func(s *clientSettings, p api.Passwords) (api.AzureProvider, error) {
+	//setupMockMSGraphClient(mockProvider)
```
austingebauer, line 48: Needed?

provider.go
```diff
@@ -5,110 +5,178 @@ package azuresecrets
...
 	if err != nil {
 		return nil, err
 	}
 
-	raClient := authorization.NewRoleAssignmentsClientWithBaseURI(settings.Environment.ResourceManagerEndpoint, settings.SubscriptionID)
-	raClient.Authorizer = resourceManagerAuthorizer
-	raClient.AddToUserAgent(userAgent)
+	p := &provider{
+		settings:   settings,
+		httpClient: httpClient,
```
austingebauer, line 93: I'm wondering if we need to hold onto this here? I see that it is given as an option for the role assignments/definitions clients. But it's unused after that point. Is there any good reason to hold onto it in a field of `provider`?

provider.go
```diff
@@ -5,110 +5,178 @@ package azuresecrets
...
-	raClient     *authorization.RoleAssignmentsClient
-	rdClient     *authorization.RoleDefinitionsClient
+	raClient     *armauthorization.RoleAssignmentsClient
+	rdClient     *armauthorization.RoleDefinitionsClient
 }
 
 // newAzureProvider creates an azureProvider, backed by Azure client objects for underlying services.
-func newAzureProvider(settings *clientSettings, passwords api.Passwords) (api.AzureProvider, error) {
+func newAzureProvider(settings *clientSettings, passwords api.Passwords) (AzureProvider, error) {
+	// @TODO see if this is still needed
```
austingebauer, line 64: It looks like we're setting the user-agent request header for the role assignments/definitions clients (via the custom transport) but not for the MS graph client. If there isn't a simple way to set the user-agent for the MS graph client, then I'd say it's okay to remove this.

Fwiw, the last we talked about these user-agent headers from MSFT they weren't making use of them for attribution/analytics.

path_rotate_root.go
```diff
@@ -8,8 +8,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/google/uuid"
```
austingebauer, line 11: Any reason for this switch in uuid library?

provider.go
```diff
@@ -137,47 +205,37 @@ func (p *provider) DeleteServicePrincipal(ctx context.Context, spObjectID string
...
+	listResp, err := page.NextPage(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	return page.Values(), nil
+	return listResp.Value, err
 }
 
 // GetRoleByID fetches the full role definition given a roleID.
```
austingebauer, line 221:
```suggestion
// GetRoleDefinitionByID fetches the full role definition given a roleID.
```

path_rotate_root.go
vinay-gopalan: Don't need this after latest refactors of internal types and methods 👍🏼

provider.go
vinay-gopalan: Updated!

provider.go
vinay-gopalan: Ah nice, I did not notice that 😅 I don't think there's any reason to tie it to the provider struct for the reasons you mentioned above 👍🏼

provider.go
vinay-gopalan: I did some preliminary research, and there doesn't seem to be clear cut documentation and recommendations on setting the User Agent for the MSGraph Client. I am seeing some threads mention a custom header could be passed on certain requests, but since it's not the official docs I'm opting to remove setting the User Agent on MSGraph client for now since it does not look like they are making use of it.

I can circle back to this in the future if it ends up that we do prefer adding in the header and if we find official recommendations from MSFT on how to set the value!

api/groups.go
```diff
@@ -62,11 +59,29 @@ func (c *MSGraphClient) ListGroups(ctx context.Context, filter string) ([]Group,
...
+
+		if groupID != nil && name != nil {
+			return Group{
+				ID:          *groupID,
+				DisplayName: *name,
+			}
+		}
+	}
+
+	// return zero-value result if app in nil
```
fairclothjm, line 81:
```suggestion
	// return zero-value result if value is nil
```

path_service_principal_test.go
```diff
@@ -764,23 +767,23 @@ func TestRoleAssignmentWALRollback(t *testing.T) {
...
 		// info from Azure and verify it matches the Reader role.
 		raIDs := credsResp.Secret.InternalData["role_assignment_ids"].([]string)
 		equal(t, 2, len(raIDs))
 
-		ra, err := provider.raClient.GetByID(context.Background(), raIDs[0])
+		ra, err := provider.raClient.GetByID(context.Background(), raIDs[0], nil)
 		assertErrorIsNil(t, err)
 
-		roleDefs, err := provider.ListRoleDefinitions(context.Background(), fmt.Sprintf("subscriptions/%s", subscriptionID), "")
+		roleDefs, err := provider.ListRoleDefinitions(nil, fmt.Sprintf("subscriptions/%s", subscriptionID), "")
```
fairclothjm, line 780: This test is failing with `unexpected error: net/http: nil Context`
```suggestion
		roleDefs, err := provider.ListRoleDefinitions(context.Background(), fmt.Sprintf("subscriptions/%s", subscriptionID), "")
```

path_service_principal_test.go
```diff
@@ -764,23 +767,23 @@ func TestRoleAssignmentWALRollback(t *testing.T) {
...
 		assertErrorIsNil(t, err)
 
-		defID := *ra.RoleAssignmentPropertiesWithScope.RoleDefinitionID
+		defID := *ra.Properties.RoleDefinitionID
 		found := false
 		for _, def := range roleDefs {
-			if *def.ID == defID && *def.RoleName == "Storage Blob Data Owner" {
+			if *def.ID == defID && *def.Name == "Storage Blob Data Owner" {
```
fairclothjm, line 786: Once we fix the nil context error this test fails with `path_service_principal_test.go:793: 'Storage Blob Data Owner' role assignment not found`
```suggestion
			if *def.ID == defID && *def.Properties.RoleName == "Storage Blob Data Owner" {
```

fairclothjm (commented):
Seeing failures:

```
=== RUN   TestRotateRootSuccess
    path_rotate_root_test.go:39: failed to add new password: Resource 'd40cee02-4c15-4396-b857-3c52894762d3' does not exist or one of its queried reference-property objects are not present.
--- FAIL: TestRotateRootSuccess (0.65s)
=== RUN   TestRotateRootPeriodicFunctionBeforeMinute
    path_rotate_root_test.go:118: failed to add new password: Resource 'd40cee02-4c15-4396-b857-3c52894762d3' does not exist or one of its queried reference-property objects are not present.
--- FAIL: TestRotateRootPeriodicFunctionBeforeMinute (0.56s)
```

api/service_principals.go
austingebauer: Thanks for the explanation, Vinay! I can see how the package structure and various client interfaces would make this difficult. I think a method that allows us to get the underlying client would work. I also think it's okay to not do this now.

path_service_principal_test.go
vinay-gopalan: Aah I thought I got all of them 😓 Thanks!

path_service_principal_test.go
vinay-gopalan: I'm surprised that the API does not also return the name one level above 🤔 Fixed now

api/service_principals.go
```diff
@@ -15,6 +20,95 @@ type ServicePrincipalClient interface {
...
+				ID:    *spID,
+				AppID: *spAppID,
+			}
+		}
+
+	}
+	return ServicePrincipal{
+		ID:    "",
+		AppID: "",
+	}
```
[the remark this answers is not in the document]
vinay-gopalan: Ah gotcha, I do think that is what the repo was doing with `autorest/to`, so this makes sense. Updated to use this approach for dereferencing string fields 👍🏼

vinay-gopalan:
> Seeing failures:
> 
> ```
> === RUN   TestRotateRootSuccess
>     path_rotate_root_test.go:39: failed to add new password: Resource 'd40cee02-4c15-4396-b857-3c52894762d3' does not exist or one of its queried reference-property objects are not present.
> --- FAIL: TestRotateRootSuccess (0.65s)
> === RUN   TestRotateRootPeriodicFunctionBeforeMinute
>     path_rotate_root_test.go:118: failed to add new password: Resource 'd40cee02-4c15-4396-b857-3c52894762d3' does not exist or one of its queried reference-property objects are not present.
> --- FAIL: TestRotateRootPeriodicFunctionBeforeMinute (0.56s)
> ```

I accidentally switched to using the `AppID` instead of the `AppObjectID` in rotate root at some point 😅 Fixed now!

client.go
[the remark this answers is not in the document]
austingebauer: Resolving this. I was thinking we'd introduce a similar helper function and seeing that we've done that now with `ptrToString` 👍

path_service_principal_test.go
```diff
@@ -1019,7 +1022,7 @@ func TestCredentialInteg_msgraph(t *testing.T) {
 		// Verify that SP get is an error after delete. Expected there
 		// to be a delay and that this step would take some time/retries,
 		// but that seems not to be the case.
-		assertServicePrincipalDoesNotExist(t, provider.spClient, spObjID)
+		assertServicePrincipalExistence(t, provider.spClient, spObjID, false)
```
austingebauer, line 1025: Quick suggestion unrelated to changes: we should check the error on `b.spRevoke` above similar to what's done in `TestRoleAssignmentWALRollback`. Will make it more obvious what failed for future test runs. Thanks!

api/applications.go
```diff
@@ -0,0 +1,229 @@
...
+		return Application{}, err
+	}
+
+	return getApplicationResponse(resp), nil
+}
+
+// DeleteApplication deletes an Azure application object.
+// This will in turn remove the service principal (but not the role assignments).
+func (c *MSGraphClient) DeleteApplication(ctx context.Context, applicationObjectID string, permanentlyDelete bool) error {
+	err := c.client.Applications().ByApplicationId(applicationObjectID).Delete(ctx, nil)
```
austingebauer, line 140: Wondering if we should check the error immediately and return it if not nil instead of combining with a multierror for the permanent delete case? I'd imagine permanent deletion will fail 100% of the time if this fails.

provider.go
```diff
@@ -25,90 +52,124 @@ type provider struct {
...
-	raClient := authorization.NewRoleAssignmentsClientWithBaseURI(settings.Environment.ResourceManagerEndpoint, settings.SubscriptionID)
-	raClient.Authorizer = resourceManagerAuthorizer
-	raClient.AddToUserAgent(userAgent)
-
-	rdClient := authorization.NewRoleDefinitionsClientWithBaseURI(settings.Environment.ResourceManagerEndpoint, settings.SubscriptionID)
-	rdClient.Authorizer = resourceManagerAuthorizer
-	rdClient.AddToUserAgent(userAgent)
-
 	p := &provider{
 		settings: settings,
```
austingebauer, line 86: Sorry another one unrelated but might be good to clean up while it's fresh 😄  We're keeping the `clientSettings` on the `azureSecretBackend`, `client`, and `provider`. I think we can safely remove it from `provider` since it's set but never used again.

austingebauer (approved):
A few small comments but otherwise LGTM! Nice job working through this one @vinay-gopalan!

path_service_principal_test.go
vinay-gopalan: Good call, added a error check on spRevoke

api/applications.go
vinay-gopalan: I like breaking early if we receive an error. Updated in latest!

provider.go
vinay-gopalan: Ah nice, removed

PR #170, opened 2023-10-30
debugging eventual consistency issue

Zlaticanin:
This PR is for debugging the eventual consistency issue and reproduce the error. 

This test sends out 100 requests in parallel to create Azure Resource Groups. The test counts how many successful requests were made compared to the failed requests due to providing an invalid secret.
This test can be ran with both app & sp credentials.

Azure [api-docs](https://developer.hashicorp.com/vault/api-docs/secret/azure) & [docs](https://developer.hashicorp.com/vault/docs/secrets/azure#roles)

1. Enable the Azure secrets engine
```
vault secrets enable azure
```
2. Configure the secrets engine with account creds (these creds can be found by logging in to Azure portal (1password) and searching for Team Vault subscription.
```
vault write azure/config \
    subscription_id=$AZURE_SUBSCRIPTION_ID \
    tenant_id=$AZURE_TENANT_ID \
    client_id=$AZURE_CLIENT_ID \
    client_secret=$AZURE_CLIENT_SECRET
```
3. To run this test with an app id, you need to register an application in the Azure portal (if you already don’t have one). 
App registrations → New registration → Register
You will need to add API permissions to your application. You will need:
       - Application.ReadWrite.All
       - GroupMember.ReadWrite.All
Make sure to Grant admin consent for Default Directory

Now go to Team Vault Subscription → Access control (IAM) → Add role assignment → Privileged administrator roles → Select Owner → Go to Members → Select Members → Add the application you just registered → Not constrained Delegation type → Review and assign

```
vault write azure/roles/test-role \
    application_object_id=<existing_app_obj_id> \
    ttl=10h
```

4. To configure a role to create a new sp with Azure roles: 
```
vault write azure/roles/test-role ttl=10h azure_roles=-<<EOF
    [
        {
            "role_name": "Contributor",
            "scope":  "/subscriptions/<uuid>"
        }
    ]
EOF
```


azure_test.go
```diff
@@ -0,0 +1,140 @@
...
+	fmt.Println("Num failures", n-int(successes))
+	fmt.Println("Num Auth failures", authFailures)
+	assert.EqualValues(t, n, successes, successes)
+
+	err = cleanup(ctx, resourceGroupClient)
+	assert.NoError(t, err)
+}
+
+func helper(ctx context.Context, rgClient *armresources.ResourceGroupsClient) (armresources.ResourceGroupsClientCreateOrUpdateResponse, error) {
+	resp, err := rgClient.CreateOrUpdate(ctx, fmt.Sprintf("%v-%v", "vault-test-{UUID}", uuid.New().String()), armresources.ResourceGroup{
```
raymonstah, line 98:
```suggestion
	resp, err := rgClient.CreateOrUpdate(ctx, fmt.Sprintf("%v-%v", "vault-test", uuid.New().String()), armresources.ResourceGroup{
```

azure_test.go
```diff
@@ -0,0 +1,140 @@
...
+		Location: to.StringPtr("West US"),
+		Tags:     map[string]*string{"created_by": to.StringPtr("vault-test-{UUID}")},
```
raymonstah, line 100: This should be an actual UUID passed in to the helper method (and the cleanup method) so that we can appropriately find the resources to delete after we're done with the test.

azure_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+	subscriptionID := os.Getenv("SUBSCRIPTION_ID")
+	tenantID := os.Getenv("TENANT_ID")
+	if subscriptionID == "" || tenantID == "" {
+		t.SkipNow()
+	}
+
+	// use dynamic credentials from Vault instead of hardcoding them.
+	// Use a regular HTTP client to make the request
+	req, err := http.NewRequest(http.MethodGet, "http://localhost:8200/v1/local-azure/creds/test-role", nil)
```
raymonstah, line 78:
```suggestion
	req, err := http.NewRequest(http.MethodGet, "http://localhost:8200/v1/azure/creds/test-role", nil)
```

azure_test.go
```diff
@@ -0,0 +1,177 @@
...
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+
+	var response VaultCredResponse
+	err = json.NewDecoder(resp.Body).Decode(&response)
+	assert.NoError(t, err)
+
+	clientID := response.Data.ClientId
+	clientSecret := response.Data.ClientSecret
+	fmt.Println("client_id: ", clientID)
+	fmt.Println("clientSecret: ", clientSecret)
```
raymonstah, line 93: can probably remove these

azure_test.go
```diff
@@ -0,0 +1,177 @@
...
+	wg.Wait()
+	fmt.Println("Num failures", n-int(successes))
+	fmt.Println("Num Auth failures", authFailures)
+	assert.EqualValues(t, n, successes, successes)
+
+	err = cleanup(ctx, resourceGroupClient)
+	assert.NoError(t, err)
+}
+
+func helper(ctx context.Context, rgClient *armresources.ResourceGroupsClient) (armresources.ResourceGroupsClientCreateOrUpdateResponse, error) {
```
raymonstah, line 134:
```suggestion
func createResouceGroup(ctx context.Context, rgClient *armresources.ResourceGroupsClient) (armresources.ResourceGroupsClientCreateOrUpdateResponse, error) {
```

azure_test.go
```diff
@@ -0,0 +1,177 @@
...
+		Location: to.StringPtr("West US"),
+		Tags:     map[string]*string{"created_by": to.StringPtr("vault-test-{UUID}")},
+	}, nil)
+	if err != nil {
+		return armresources.ResourceGroupsClientCreateOrUpdateResponse{}, err
+	}
+	return resp, nil
+}
+
+func cleanup(ctx context.Context, rgClient *armresources.ResourceGroupsClient) error {
```
raymonstah, line 145:
```suggestion
func cleanupResourceGroups(ctx context.Context, rgClient *armresources.ResourceGroupsClient) error {
```

go.mod
```diff
@@ -1,35 +1,44 @@
...
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.8.0
+	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.4.0
+	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources v1.1.1
 	github.com/Azure/go-autorest/autorest v0.11.29
 	github.com/Azure/go-autorest/autorest/azure/auth v0.5.12
 	github.com/Azure/go-autorest/autorest/date v0.3.0
 	github.com/Azure/go-autorest/autorest/to v0.4.0
 	github.com/go-test/deep v1.1.0
 	github.com/golang/mock v1.6.0
+	github.com/google/uuid v1.3.1
```
raymonstah, line 16: ah, didn't realize we'd be importing this library. Can we try to use `go-uuid` instead?

go.mod
```diff
@@ -1,9 +1,12 @@
 module github.com/hashicorp/vault-plugin-secrets-azure
 
-go 1.19
+go 1.18
```
raymonstah, line 3: Should this be downgraded?

azure_test.go
```diff
@@ -0,0 +1,176 @@
...
+					}
+				}
+				return err
+			})
+		}
+		if err := group.Wait(); err != nil {
+			return err
+		}
+	}
+	fmt.Println("items marked for deletion", counter)
```
raymonstah, line 174: Maybe we can use `t.Logf` instead of `fmt` here and everywhere else we call Println

raymonstah (approved):

go.mod
Zlaticanin: I had to downgrade it to run go mod tidy, will switch it back 👍

raymonstah:
I think it's safe to close this PR at this point. This was instrumental in getting us the data points we needed to provide to Microsoft, and may be revisited when a fix is implemented.

PR #173, opened 2023-11-13
bootstrap: output application object ID for testing

fairclothjm:
This PR updates the terraform bootstrap setup to output the Application Object ID that can be used in testing existing service principle credential generation.


austingebauer (approved):
👍

PR #174, opened 2023-11-21
Azure Secrets Engine Customizations

Zlaticanin:
Adding two configurable fields `sign_in_audience` and `tags` to application registration. 

To test:

```
vault secrets enable azure

vault write azure/config \
    subscription_id=$AZURE_SUBSCRIPTION_ID \
    tenant_id=$AZURE_TENANT_ID \
    client_id=$AZURE_CLIENT_ID \
    client_secret=$AZURE_CLIENT_SECRET

vault write azure/roles/my-role \      
    ttl=1h \
    max_ttl=24h \
    azure_roles=@az_roles.json \
    sign_in_audience=AzureADMyOrg \
    tags="team:engineering","environment:development"

vault read azure/creds/my-role
```

Use the `client_id` to search through all [app registrations in Azure portal](https://portal.azure.com/#view/Microsoft_AAD_RegisteredApps/ApplicationsListBlade) to check that the fields are correctly set.

### DIFF
--- a/api/applications.go
+++ b/api/applications.go
@@ -124,8 +124,12 @@ func (c *MSGraphClient) ListApplications(ctx context.Context, filter string) ([]
 func (c *MSGraphClient) CreateApplication(ctx context.Context, displayName string, signInAudience string, tags []string) (Application, error) {
 	requestBody := models.NewApplication()
 	requestBody.SetDisplayName(&displayName)
-	requestBody.SetSignInAudience(&signInAudience)
 	requestBody.SetTags(tags)
+
+	// only set signInAudience if it's non-empty
+	if signInAudience != "" {
+		requestBody.SetSignInAudience(&signInAudience)
+	}
 
 	resp, err := c.client.Applications().Post(ctx, requestBody, nil)
 	if err != nil {
--- a/api/applications.go
+++ b/api/applications.go
@@ -18,7 +18,7 @@ import (
 
 type ApplicationsClient interface {
 	GetApplication(ctx context.Context, clientID string) (Application, error)
-	CreateApplication(ctx context.Context, displayName string) (Application, error)
+	CreateApplication(ctx context.Context, displayName string, signInAudience string, tags []string) (Application, error)
 	DeleteApplication(ctx context.Context, applicationObjectID string, permanentlyDelete bool) error
 	ListApplications(ctx context.Context, filter string) ([]Application, error)
 	AddApplicationPassword(ctx context.Context, applicationObjectID string, displayName string, endDateTime time.Time) (PasswordCredential, error)
@@ -121,9 +121,11 @@ func (c *MSGraphClient) ListApplications(ctx context.Context, filter string) ([]
 }
 
 // CreateApplication create a new Azure application object.
-func (c *MSGraphClient) CreateApplication(ctx context.Context, displayName string) (Application, error) {
+func (c *MSGraphClient) CreateApplication(ctx context.Context, displayName string, signInAudience string, tags []string) (Application, error) {
 	requestBody := models.NewApplication()
 	requestBody.SetDisplayName(&displayName)
+	requestBody.SetSignInAudience(&signInAudience)
+	requestBody.SetTags(tags)
 
 	resp, err := c.client.Applications().Post(ctx, requestBody, nil)
 	if err != nil {
--- a/path_roles.go
+++ b/path_roles.go
@@ -13,6 +13,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization/v2"
 	"github.com/hashicorp/vault/sdk/framework"
 	"github.com/hashicorp/vault/sdk/helper/jsonutil"
+	"github.com/hashicorp/vault/sdk/helper/strutil"
 	"github.com/hashicorp/vault/sdk/logical"
 
 	"github.com/hashicorp/vault-plugin-secrets-azure/api"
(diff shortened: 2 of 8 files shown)


path_roles.go
```diff
@@ -87,6 +89,14 @@ func pathsRole(b *azureSecretBackend) []*framework.Path {
...
 					Description: "JSON list of Azure groups to add the service principal to.",
 				},
+				"sign_in_audience": {
+					Type:        framework.TypeString,
+					Description: "Specifies the security principal types that are allowed to sign in to the application.",
+				},
+				"tags": {
+					Type:        framework.TypeCommaStringSlice,
+					Description: "Azure tags to attach to an application.",
+				},
```
maxcoulombe, line 99: It seems like non-KV formats work, but to be sure we are fine if tags do not exactly follow a `"key:value"` format?
```
vault write azure/roles/my-role ... tags="just-a-string"
vault write azure/roles/my-role ... tags="not-a-colon=delimiter"
etc.
```
Are working just fine. If this is on purpose it should probably be included in tests so clearly signal to future maintainers that this is on purpose and not a missing validation.

path_roles_test.go
```diff
@@ -217,6 +223,8 @@ func TestRoleCreate(t *testing.T) {
 				}]`,
 			),
 			"application_object_id": "",
+			"sign_in_audience":      "",
+			"tags":                  []string{""},
 			"azure_groups":          "[]",
 			"persist_app":           false,
 		}
```
maxcoulombe, line 230: Should we add coverage for some of the invalid cases to confirm the plugin behaves in a predictable way? For example, Azure tags are case-insensitive so creating a role with:
`vault write azure/roles/my-role-1 ... sign_in_audience=AzureADMyOrg tags="case","CASE"`

Results in this error when generating credentials:
```
Error reading local-secrets-azure/creds/my-role-1: Error making API request.

URL: GET http://127.0.0.1:8200/v1/local-secrets-azure/creds/my-role-1
Code: 500. Errors:

* 1 error occurred:
	* Request contains a property with duplicate values.
```

I know there are precedents in Vault that we sometimes prefer failing with a 500 instead of adding pre-conditions to return a client error. But it should be tested or documented so if we decide to let the Azure errors bubble up instead of validating the inputs it is deliberate.

maxcoulombe (commented):
Looks good! I mostly have open questions about UX/validations/backward-compatibility.

path_roles.go
```diff
@@ -208,6 +218,19 @@ func (b *azureSecretBackend) pathRoleUpdate(ctx context.Context, req *logical.Re
 		role.PermanentlyDelete = false
 	}
 
+	// update and verify SignInAudience if provided
+	if signInAudience, ok := d.GetOk("sign_in_audience"); ok {
+		role.SignInAudience = signInAudience.(string)
+	}
```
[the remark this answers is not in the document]
Zlaticanin: I just added some validation to check if the audience provided is valid, if not it errors out with the reason why

api/applications.go
```diff
@@ -121,9 +121,11 @@ func (c *MSGraphClient) ListApplications(ctx context.Context, filter string) ([]
 }
 
 // CreateApplication create a new Azure application object.
-func (c *MSGraphClient) CreateApplication(ctx context.Context, displayName string) (Application, error) {
+func (c *MSGraphClient) CreateApplication(ctx context.Context, displayName string, signInAudience string, tags []string) (Application, error) {
 	requestBody := models.NewApplication()
 	requestBody.SetDisplayName(&displayName)
+	requestBody.SetSignInAudience(&signInAudience)
+	requestBody.SetTags(tags)
```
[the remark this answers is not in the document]
Zlaticanin: Thank you! Should be fixed now

path_roles.go
Zlaticanin: This is intentional, for app registrations Azure only allows tags to be passed as an array of strings unfortunately. 
https://feedback.azure.com/d365community/idea/ec2330ed-b125-ec11-b6e6-000d3a4f0789

path_roles_test.go
Zlaticanin: Good call. I added validation based on the docs here: https://learn.microsoft.com/en-us/entra/identity-platform/supported-accounts-validation

I also added some test cases for invalid tags and sign in audience to make sure it behaves correctly and we do get an error

api/applications.go
```diff
@@ -18,7 +18,7 @@ import (
 
 type ApplicationsClient interface {
 	GetApplication(ctx context.Context, clientID string) (Application, error)
-	CreateApplication(ctx context.Context, displayName string) (Application, error)
+	CreateApplication(ctx context.Context, displayName string, signInAudience string, tags []string) (Application, error)
```
raymonstah, line 21: minor, perhaps worth a followup PR:
It might be worthwhile introducing a struct to hold the bag of parameters.
I typically use the convention of `*Input`, so in this case it would be:
```
type CreateApplicationInput struct {
...
```

Since this appears to be an exported interface, we don't want to break clients every time we add a new field, causing them to make updates to match the new signature.

path_roles.go
```diff
@@ -208,6 +218,53 @@ func (b *azureSecretBackend) pathRoleUpdate(ctx context.Context, req *logical.Re
...
+		}
+		if !valid {
+			validValuesString := strings.Join(validSignInAudiences, ", ")
+			return logical.ErrorResponse("Invalid value for sign_in_audience field. Valid values are: %s", validValuesString), nil
+		}
+		role.SignInAudience = signInAudienceValue
+	}
+
+	// update and validate Tags if provided
+	if tags, ok := d.GetOk("tags"); ok {
```
raymonstah, line 244: If we extract this out to a separate `validateTags` function I think it would make for some clean table tests.

raymonstah (approved):
nice work!

path_roles.go
```diff
@@ -208,6 +218,53 @@ func (b *azureSecretBackend) pathRoleUpdate(ctx context.Context, req *logical.Re
...
+		for _, validValue := range validSignInAudiences {
+			if signInAudienceValue == validValue {
+				valid = true
+				break
+			}
+		}
+		if !valid {
+			validValuesString := strings.Join(validSignInAudiences, ", ")
+			return logical.ErrorResponse("Invalid value for sign_in_audience field. Valid values are: %s", validValuesString), nil
+		}
```
austingebauer, line 239: The Vault SDK `strutil` package has some nice helpers for this type of validation. Might consider simplifying to something like:
```go
validSignInAudiences := []string{"AzureADMyOrg", "AzureADMultipleOrgs", "AzureADandPersonalMicrosoftAccount", "PersonalMicrosoftAccount"}
if !strutil.StrListContains(validSignInAudiences, signInAudienceValue) {
	validValuesString := strings.Join(validSignInAudiences, ", ")
	return logical.ErrorResponse("Invalid value for sign_in_audience field. Valid values are: %s", validValuesString), nil
}
```

path_roles.go
```diff
@@ -208,6 +218,53 @@ func (b *azureSecretBackend) pathRoleUpdate(ctx context.Context, req *logical.Re
...
+					return logical.ErrorResponse("individual tag size must be between 1 and 256 characters (inclusive)"), nil
+				}
+				// Check for whitespaces
+				if strings.Contains(tag, " ") {
+					return logical.ErrorResponse("whitespaces are not allowed in tags"), nil
+				}
+				// Check for duplicates
+				if _, exists := uniqueTags[tag]; exists {
+					return logical.ErrorResponse("duplicate tags are not allowed"), nil
+				}
```
austingebauer, line 259: I'm wondering if it'd be a little more forgiving and user-friendly to trim+deduplicate the tags for users? Some Vault APIs do this using [`RemoveDuplicates`](https://github.com/hashicorp/go-secure-stdlib/blob/main/strutil/strutil.go#L235).

austingebauer (approved):
A couple of minor suggestions but otherwise LGTM!

path_roles.go
Zlaticanin: Nice! Thank you. I added this

path_roles.go
Zlaticanin: Thank you! Thank looks a lot nicer, updated in the last commit!

api/applications.go
Zlaticanin: Sounds good, I will give this a shot in a followup

PR #181, opened 2024-02-23
Deprecate password policy

Zlaticanin:


path_service_principal_test.go
```diff
@@ -1044,8 +1048,8 @@ func assertClientSecret(tb testing.TB, data map[string]interface{}) {
 	if !ok {
 		tb.Fatalf("client_secret is not a string")
 	}
-	if len(actualPassword) != api.PasswordLength {
-		tb.Fatalf("client_secret is not the correct length: expected %d but was %d", api.PasswordLength, len(actualPassword))
+	if len(actualPassword) != PasswordLength {
+		tb.Fatalf("client_secret is not the correct length: expected %d but was %d", PasswordLength, len(actualPassword))
 	}
```
raymonstah, line 1053: What's setting the password in this scenario? How / why is the password length still 36?

raymonstah (approved):
Just had a question out of curiosity regarding the password length in the test. It might be a mock / stub, but wanted to clarify!

path_service_principal_test.go
Zlaticanin: I left this because of [this](https://github.com/hashicorp/vault-plugin-secrets-azure/blob/main/backend_test.go#L30) mock. Tbh, I am not really sure why the correct length is 36.

austingebauer: I think it's a 36 character UUID from the mock at https://github.com/hashicorp/vault-plugin-secrets-azure/blob/main/provider_mock_test.go#L176.

austingebauer (approved):
Thanks, @Zlaticanin!

PR #186, opened 2024-04-03
Changelog++ for v0.18.0

tomhjp:
Prepare for v0.17.1 release which will be published to the releases website too.

The dep updates are to satisfy the security scanner:

```shell-session
$ scan binary bin/vault-plugin-secrets-azure
✓ Scanned file:{path:"bin/vault-plugin-secrets-azure"} in 14.5s - no results found
```


tomhjp:
I missed #181 when I was checking for changelog updates - perhaps this should be v0.18.0 given the change? @MilenaHC are you happy with the changelog notice I've written for that?

Zlaticanin (approved):
LGTM! Sorry about the late approval, just now saw this.

vinay-gopalan (changes requested):
Can we add in a line for the bug fix similar to #192? Thanks!

vinay-gopalan (approved):

PR #188, opened 2024-04-12
plugin workload identity

fairclothjm:
This PR adds plugin WIF support to the Azure secrets engine. The following new fields to the config endpoint to enable configuring Workload Identity Federation:
- identity_token_audience
- identity_token_ttl


vinay-gopalan (approved):
LGTM! Great work 🎉

PR #196, opened 2024-04-24
correct msgraph search query - fixes #189

### LINKED ISSUE #189: Getting InternalServerErrors returned when creating new roles in the engine
Recently we upgraded the azure secrets engine from 0.16.1 to 0.17.0 when we upgraded our Vault from 1.14.8 to 1.16.1.
It seems there is now an issue when creating new roles:

```
vault write azure/engine/roles/role-name \
>     application_object_id=#### \
>     ttl=1h
Error writing data to azure/engine/roles/role-name: Error making API request.

URL: PUT https://****.vault.****.com/v1/azure/engine/roles/role-name
Code: 500. Errors:

* 1 error occurred:
        * error loading Application: no application found
```

I checked the validity of the token that's in the config for the engine, it's valid and working. Since the error is 'no application found'; I tested it by using the application id, subscription id, tenant id and client secret in the config and putting it in my Postman, and performing a GET API call like so:
```
GET {{graphurl}}/applications/####
```
I also tested the creation of a new application, and a secret on the application, using Postman: Those also work.
I tested the generation of a secret via the azure secrets engine: Works.
So there is nothing wrong with permissions on the identity used in the config. The issue seems to be scoped to just the creation of roles into the secrets engine.

The InternalServerError seems to be coming from the engine itself, or from Vault maybe, since it fails on doing the PUT command, but then is giving an error that suggests the issue is on the AAD side.

I also looked into the Vault logs, however those do not provide any more relevant information then is already listed here.

sspans-sbp:
# Overview

The refactored validation for static service principals was checking the **application_object_id** by searching for an **application_id**. This updates the search to search for **application_object_id** instead.

# Design of Change
Only the search filter was updated.  
The lookup could possibly be improved by fetching the application object directly rather than searching. But in this case I've opted for the smallest change to resolve the regression.

[x] [Issue #189](https://github.com/hashicorp/vault-plugin-secrets-azure/issues/189)

# Contributor Checklist
[My Docs PR Link](link)
[x] Backwards compatible

### DIFF
--- a/api/applications.go
+++ b/api/applications.go
@@ -73,8 +73,8 @@ func NewMSGraphClient(graphURI string, creds azcore.TokenCredential) (*MSGraphCl
 	return ac, nil
 }
 
-func (c *MSGraphClient) GetApplication(ctx context.Context, clientID string) (Application, error) {
-	filter := fmt.Sprintf("appId eq '%s'", clientID)
+func (c *MSGraphClient) GetApplication(ctx context.Context, applicationObjectID string) (Application, error) {
+	filter := fmt.Sprintf("Id eq '%s'", applicationObjectID)
 	req := applications.ApplicationsRequestBuilderGetRequestConfiguration{
 		QueryParameters: &applications.ApplicationsRequestBuilderGetQueryParameters{
 			Filter: &filter,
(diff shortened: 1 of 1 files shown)


sspans-sbp:
@vinay-gopalan any chance you could review / merge this one?

gsantos-hc:
@sspans-sbp I found the same bug while working on something else. Agree that it's only in Vault 1.16+.

There's also a `GetApplication` interface at the top of `api/applications.go`. The interface's variable name is still `clientId`, which will be confusing in IDEs. Would you also update that to `objectId`?

sspans-sbp:
A better solution was merged in #200

PR #198, opened 2024-05-01
fix(api): get apps by object id

gsantos-hc:
# Overview

Fix an inconsistency between the `GetApplication` function's behavior and that of its callers.

All callers to the function pass in an application's Object ID. However, the function expected the Client ID.

This led to inconsistent behaviors when configuring a role with `application_object_id`:
- You had to pass in the Client ID for the role to be configured successfully in the first place, but...
- Then you couldn't create secrets for that application because the function handling secret creation expects the Object ID.

# Design of Change

Detected this bug while implementing a separate feature. To reproduce, try configuring an Azure role with `application_object_id`. You _shouldn't_ be able to configure the role using the Object ID. If you use the Client ID instead, you can configure the role but then you can't create secrets for that application.

# Contributor Checklist
[My Docs PR Link](link)
[ ] Backwards compatible

### DIFF
--- a/api/applications.go
+++ b/api/applications.go
@@ -73,8 +73,8 @@ func NewMSGraphClient(graphURI string, creds azcore.TokenCredential) (*MSGraphCl
 	return ac, nil
 }
 
-func (c *MSGraphClient) GetApplication(ctx context.Context, clientID string) (Application, error) {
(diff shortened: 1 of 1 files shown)


gsantos-hc:
Looks like there was already another fix in flight for this. Closing in favor of #196 .

PR #200, opened 2024-05-02
Correctly use GetApplication with `ApplicationObjectID` instead of incorrect filter

vinay-gopalan:
# Overview
When performing the MSGraph SDK upgrade, the `GetApplication` method was mistranslated. It was expecting a `clientID` to be passed in, and it would then filter off of the client ID in order to fetch the Application. However, the actual usage of `GetApplication` is instead meant to use the `ApplicationObjectID` to query for the application. This can be seen [on this line here](https://github.com/hashicorp/vault-plugin-secrets-azure/blob/main/path_roles.go#L254), where `GetApplication` is only used when the `ApplicationObjectID` is a non-zero value. 

This PR fixes the mistranslation, and correctly fetches the application based on the `ApplicationObjectID` instead of the Client ID.

# Contributor Checklist
[My Docs PR Link](link)
[ ] Backwards compatible

### DIFF
--- a/path_roles_test.go
+++ b/path_roles_test.go
@@ -6,11 +6,14 @@ package azuresecrets
 import (
 	"context"
 	"fmt"
+	"os"
 	"sort"
 	"strings"
 	"testing"
 	"time"
 
+	log "github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/vault/sdk/helper/logging"
 	"github.com/hashicorp/vault/sdk/logical"
 )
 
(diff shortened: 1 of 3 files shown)


gsantos-hc:
As noted in #196: it'd be good to also rename `clientId` to `objectId` in the interface at the top of the file

path_roles_test.go
```diff
@@ -557,6 +560,89 @@ func TestRoleCreateBad(t *testing.T) {
 	}
 }
 
+// TestRolesCreate_applicationObjectID tests that a role
+// cna be created using the Application Object ID field.
```
Zlaticanin, line 564:
```suggestion
// can be created using the Application Object ID field.
```

Zlaticanin (approved):
LGTM! Thank you for fixing this 🙏

fairclothjm (approved):

PR #210, opened 2024-06-11
Correct nil check when unassigning roles

jmarion-arista:
This is a continued fix from hashicorp/vault-plugin-secrets-azure#191 (reported in hashicorp/vault-plugin-secrets-azure#190)

The issue with the original fix is that Go will still try to evaluate the right side of the `||` operation and try to dereference `.StatusCode` of a `nil` `rawResponse`.

Here is a Go playground PoC whipped up by my colleague: https://go.dev/play/p/tu6hOZgv7ZT

# Fixes
hashicorp/vault-plugin-secrets-azure#190

PR #213, opened 2024-06-27
Fix segmentation fault when unassigning role assignments

vinay-gopalan:
# Overview
The error handling case in `unassignRoles` was misconstrued, and was causing a panic when unassigning role assignments would fail during a WAL Rollback. This PR appropriately parses the received error from Azure and fixes the if cases to remove the panic case.

A unit test with a mock provider was added to confirm that the plugin can correctly
- return errors when needed
- ignore errors if the status code is 204 or 404 (role assignment was manually deleted)

[x] [Issue #190]
[x] [PR #210]

# Contributor Checklist
[My Docs PR Link](link)
[x] Backwards compatible

test output without fix:
```
=== RUN   TestUnassignRoleFailures
=== RUN   TestUnassignRoleFailures/Role_unassign_fail
--- FAIL: TestUnassignRoleFailures (0.00s)
    --- FAIL: TestUnassignRoleFailures/Role_unassign_fail (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x3b4b916]
```

test output with fix:
```
=== RUN   TestUnassignRoleFailures
=== RUN   TestUnassignRoleFailures/Role_unassign_fail
=== RUN   TestUnassignRoleFailures/Role_unassign_error_handled
--- PASS: TestUnassignRoleFailures (0.00s)
    --- PASS: TestUnassignRoleFailures/Role_unassign_fail (0.00s)
    --- PASS: TestUnassignRoleFailures/Role_unassign_error_handled (0.00s)
PASS
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -214,10 +214,10 @@ func (c *client) unassignRoles(ctx context.Context, roleIDs []string) error {
 	var merr *multierror.Error
 
 	for _, id := range roleIDs {
-		detailedErr := new(azcore.ResponseError)
 		if _, err := c.provider.DeleteRoleAssignmentByID(ctx, id); err != nil {
-			// If a role was deleted manually then Azure returns a error and status 204
-			if errors.As(err, &detailedErr) && (detailedErr.StatusCode == http.StatusNoContent || detailedErr.StatusCode == http.StatusNotFound) {
+			// If a role was deleted manually then Azure returns an error and status 204
+			respErr := new(azcore.ResponseError)
+			if errors.As(err, &respErr) && (respErr.StatusCode == http.StatusNoContent || respErr.StatusCode == http.StatusNotFound) {
 				continue
 			}
 
--- a/client.go
+++ b/client.go
@@ -215,7 +215,7 @@ func (c *client) unassignRoles(ctx context.Context, roleIDs []string) error {
 
 	for _, id := range roleIDs {
 		if _, err := c.provider.DeleteRoleAssignmentByID(ctx, id); err != nil {
-			// If a role was deleted manually then Azure returns an error and status 204
+			// If a role was deleted out-of-band then Azure returns an error and status 204
 			respErr := new(azcore.ResponseError)
 			if errors.As(err, &respErr) && (respErr.StatusCode == http.StatusNoContent || respErr.StatusCode == http.StatusNotFound) {
 				continue
--- a/client.go
+++ b/client.go
@@ -13,8 +13,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization/v2"
 	"github.com/google/uuid"
 	"github.com/hashicorp/go-multierror"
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -214,11 +214,10 @@ func (c *client) unassignRoles(ctx context.Context, roleIDs []string) error {
 	var merr *multierror.Error
 
 	for _, id := range roleIDs {
-		var rawResponse *http.Response
-		ctxWithResp := policy.WithCaptureResponse(ctx, &rawResponse)
-		if _, err := c.provider.DeleteRoleAssignmentByID(ctxWithResp, id); err != nil {
+		detailedErr := new(azcore.ResponseError)
+		if _, err := c.provider.DeleteRoleAssignmentByID(ctx, id); err != nil {
```
benashz, line 218: nit: we only need to create the response err in the case where an error has occurred. We can probably rename the var to `respError` since there appears to be no other class of error returned by the deletion call.
```suggestion
		if _, err := c.provider.DeleteRoleAssignmentByID(ctx, id); err != nil {
				detailedErr := new(azcore.ResponseError)
```

client.go
```diff
@@ -214,11 +214,10 @@ func (c *client) unassignRoles(ctx context.Context, roleIDs []string) error {
...
 			// If a role was deleted manually then Azure returns a error and status 204
```
benashz, line 219:
```suggestion
			// If a role was deleted manually then Azure returns an error and status 204
```

client.go
```diff
@@ -214,11 +214,10 @@ func (c *client) unassignRoles(ctx context.Context, roleIDs []string) error {
...
-			if rawResponse != nil && rawResponse.StatusCode == http.StatusNoContent || rawResponse.StatusCode == http.StatusNotFound {
+			if errors.As(err, &detailedErr) && (detailedErr.StatusCode == http.StatusNoContent || detailedErr.StatusCode == http.StatusNotFound) {
```
benashz, line 220:
```suggestion
			if errors.As(err, &respErr) && (detailedErr.StatusCode == http.StatusNoContent || respErr.StatusCode == http.StatusNotFound) {
```

benashz, line 220: It seems we are only testing the 204 HTTP status. Could we also test the 404 case?

benashz (commented):
Looks good. The approach to the evaluate there response error was definitely the right way to go.

client.go
vinay-gopalan: Added a test step for the 404 case in [8053ffa](https://github.com/hashicorp/vault-plugin-secrets-azure/pull/213/commits/8053ffa1661044488d419bf15ae1466cc6f8a4fa)

client.go
```diff
@@ -214,11 +214,10 @@ func (c *client) unassignRoles(ctx context.Context, roleIDs []string) error {
 	var merr *multierror.Error
 
 	for _, id := range roleIDs {
-		var rawResponse *http.Response
-		ctxWithResp := policy.WithCaptureResponse(ctx, &rawResponse)
-		if _, err := c.provider.DeleteRoleAssignmentByID(ctxWithResp, id); err != nil {
-			// If a role was deleted manually then Azure returns a error and status 204
-			if rawResponse != nil && rawResponse.StatusCode == http.StatusNoContent || rawResponse.StatusCode == http.StatusNotFound {
+		if _, err := c.provider.DeleteRoleAssignmentByID(ctx, id); err != nil {
+			// If a role was deleted manually then Azure returns an error and status 204
```
fairclothjm, line 218: Does manually mean out-of-band, i.e. not by the plugin?

provider_mock_test.go
```diff
@@ -224,6 +232,21 @@ func (m *mockProvider) CreateRoleAssignment(_ context.Context, scope string, nam
 }
 
 func (m *mockProvider) DeleteRoleAssignmentByID(_ context.Context, _ string) (armauthorization.RoleAssignmentsClientDeleteByIDResponse, error) {
+	if m.failUnassignRoles {
+		if m.unassignRolesFailureParams.expectError {
+			// return empty response and no 200 status codes to throw error
+			return armauthorization.RoleAssignmentsClientDeleteByIDResponse{}, &azcore.ResponseError{
+				ErrorCode: "mock: fail to delete role assignment",
+			}
+		} else {
```
fairclothjm, line 241: nit: could remove this else

fairclothjm (approved):
LGTM!

client.go
vinay-gopalan: That is correct. I updated the comment to reflect that it is an 'out-of-band' change to be more clear

provider_mock_test.go
vinay-gopalan: Good catch, removed :)

PR #215, opened 2024-07-01
Upgrade dependencies for v0.19.2 release

vinay-gopalan:
Upgrade dependencies to make Vault security scanner happy.


fairclothjm (approved):
LGTM, is the title of the PR incorrect? Should be 0.19.2 right?

vinay-gopalan:
> LGTM, is the title of the PR incorrect? Should be 0.19.2 right?

That's correct, thanks for the catch! Updated :)

PR #227, opened 2025-01-30
Implementation of scheduled root rotation (enterprise only)

kpcraig:


path_rotate_root.go
```diff
@@ -36,14 +37,28 @@ func pathRotateRoot(b *azureSecretBackend) *framework.Path {
...
+	// this replicates the old case where a wal error returned the multierror and a blank response instead of nil.
+	var error *multierror.Error
+	if errors.As(err, &error) {
+		return &logical.Response{}, err
+	}
+
+	return nil, err
+}
+
+func (b *azureSecretBackend) rotateCredential(ctx context.Context, req *logical.Request) error {
```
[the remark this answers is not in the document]
kpcraig: in the other PRs i called it rotateRootCredential for this reason, so i'll change it here

path_config.go
```diff
@@ -188,6 +209,36 @@ func (b *azureSecretBackend) pathConfigWrite(ctx context.Context, req *logical.R
...
+			RotationSchedule: config.RotationSchedule,
+			RotationWindow:   config.RotationWindow,
+			RotationPeriod:   config.RotationPeriod,
+		}
+
+		_, err := b.System().RegisterRotationJob(ctx, req)
+		if err != nil {
+			return logical.ErrorResponse("error registering rotation job: %s", err), nil
+		}
+	}
```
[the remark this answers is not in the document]
kpcraig: I updated this to match what we're doing in `logical/databases` - do the rotation registration first, then the config update, and if the config update errors, add a new warning to the resulting message that there is a dangling rotation registration.

kpcraig:
For testing, we've been using plugin-side testing mostly for validation, and relying on the vault-side to test path into the rotation manager, and that calls out are made to the correct endpoints.

path_config.go
```diff
@@ -188,9 +210,47 @@ func (b *azureSecretBackend) pathConfigWrite(ctx context.Context, req *logical.R
...
+		if prevRotateCfg {
+			err := b.System().DeregisterRotationJob(ctx, deregisterReq)
+			if err != nil {
+				return logical.ErrorResponse("error de-registering rotation job: %s", err), nil
+			}
+		}
+	} else if config.ShouldRegisterRotationJob() {
+		rotOp = "registeration"
+		req := &rotation.RotationJobConfigureRequest{
+			Name:             rootRotationJobName,
```
vinay-gopalan, line 232: nit: can skip this

path_config.go
```diff
@@ -188,9 +210,47 @@ func (b *azureSecretBackend) pathConfigWrite(ctx context.Context, req *logical.R
...
+			ReqPath:    req.Path,
+		}
+		if prevRotateCfg {
+			err := b.System().DeregisterRotationJob(ctx, deregisterReq)
+			if err != nil {
+				return logical.ErrorResponse("error de-registering rotation job: %s", err), nil
+			}
+		}
+	} else if config.ShouldRegisterRotationJob() {
+		rotOp = "registeration"
```
vinay-gopalan, line 230: nit: can use constants now that they are available

path_config.go
```diff
@@ -188,9 +210,47 @@ func (b *azureSecretBackend) pathConfigWrite(ctx context.Context, req *logical.R
...
+		_, err := b.System().RegisterRotationJob(ctx, req)
+		if err != nil {
+			return logical.ErrorResponse("error registering rotation job: %s", err), nil
+		}
+	}
+
 	err = b.saveConfig(ctx, config, req.Storage)
 	if err != nil {
-		return nil, err
+		wrappedError := err
```
vinay-gopalan, line 248: can we add the log back in since we're doing both the log and the error in the other plugins?

path_config.go
```diff
@@ -188,9 +210,47 @@ func (b *azureSecretBackend) pathConfigWrite(ctx context.Context, req *logical.R
...
+	var rotOp string
+	if config.ShouldDeregisterRotationJob() {
+		rotOp = "deregistration"
+		// Ensure de-registering only occurs on updates and if
+		// a credential has actually been registered (rotation_period or rotation_schedule is set)
+		deregisterReq := &rotation.RotationJobDeregisterRequest{
+			MountPoint: req.MountPoint,
+			ReqPath:    req.Path,
+		}
+		if prevRotateCfg {
```
vinay-gopalan, line 223: Don't think this is a blocker, but this may be inconsistent with AWS and DB currently, since they do not check for the previous config. I don't think it is wrong to do so though, just depends on what we deem to be the best

path_config_test.go
```diff
@@ -204,19 +224,23 @@ func TestConfigDelete(t *testing.T) {
...
 	config := map[string]interface{}{
-		"subscription_id":         "a228ceec-bf1a-4411-9f95-39678d8cdb34",
-		"tenant_id":               "7ac36e27-80fc-4209-a453-e8ad83dc18c2",
-		"client_id":               "testClientId",
-		"client_secret":           "testClientSecret",
-		"environment":             "AZURECHINACLOUD",
-		"root_password_ttl":       int((24 * time.Hour).Seconds()),
-		"identity_token_audience": "",
-		"identity_token_ttl":      int64(0),
+		"subscription_id":            "a228ceec-bf1a-4411-9f95-39678d8cdb34",
```
vinay-gopalan, line 227: nit: might be worth adding the CE error path test case for Azure similar to [AWS](https://github.com/hashicorp/vault/blob/main/builtin/logical/aws/path_config_root_test.go#L269)

kpcraig: added

path_config.go
kpcraig: removed

vinay-gopalan (dismissed):

path_config.go
```diff
@@ -188,9 +204,49 @@ func (b *azureSecretBackend) pathConfigWrite(ctx context.Context, req *logical.R
...
+			ReqPath:    req.Path,
+		}
+
+		err := b.System().DeregisterRotationJob(ctx, deregisterReq)
+		if err != nil {
+			return logical.ErrorResponse("error de-registering rotation job: %s", err), nil
+		}
+
+	} else if config.ShouldRegisterRotationJob() {
+		rotOp = "registeration"
```
fairclothjm, line 224: should we use the constants?

path_config.go
```diff
@@ -188,9 +204,49 @@ func (b *azureSecretBackend) pathConfigWrite(ctx context.Context, req *logical.R
...
+		req := &rotation.RotationJobConfigureRequest{
+			Name:             rootRotationJobName,
```
fairclothjm, line 226: Can we remove this?

path_config.go
```diff
@@ -21,13 +24,17 @@ const (
 	// the Azure UI, so we're setting it to 6 months (in hours)
 	// as the default.
 	defaultRootPasswordTTL = 4380 * time.Hour
+
+	// name to use on rotation jobs submitted to the rotation manager
+	rootRotationJobName = "azure-secrets-root-creds"
```
fairclothjm, line 29: can we remove this?

fairclothjm (commented):
Should we add a changelog entry?

kpcraig:
Added a changelog entry

path_config.go
kpcraig: We can use the constants but it requires us to move off the sdk tag. The values between plugins don't have to be the same for any technical reason although I agree this should become the constants at some point.

path_config.go
kpcraig: removed

path_config.go
kpcraig: removed

fairclothjm (approved):

PR #245, opened 2025-03-26
Forward Performance Secondary Requests when configuring root credentials

vinay-gopalan:
Fixes a bug where Performance Secondaries panic when they attempt to de-register / register jobs to the RM, since the RM is nil on Perf Secondaries. This fix appropriately forwards such requests to the Primary node in the cluster.

### DIFF
--- a/path_config.go
+++ b/path_config.go
@@ -98,12 +98,16 @@ func pathConfig(b *azureSecretBackend) *framework.Path {
 				DisplayAttrs: &framework.DisplayAttributes{
 					OperationVerb: "configure",
 				},
+				ForwardPerformanceSecondary: true,
+				ForwardPerformanceStandby:   true,
 			},
 			logical.UpdateOperation: &framework.PathOperation{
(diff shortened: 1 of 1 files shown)


kpcraig (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## Unreleased
+
+BUG FIXES:
```
fairclothjm, line 3: Should we go ahead and add the version title so we don't have to open another PR for that?

kpcraig:
There's an open PR (https://github.com/hashicorp/vault-plugin-secrets-azure/pull/240) that just adds a changelog - i assume that means the change described there is already in, so i imagine we also want that pulled in before we cut a release?

CHANGELOG.md
vinay-gopalan: Added next release version

fairclothjm (approved):

PR #249, opened 2025-05-22
Fix spRenew bug related to missing key_end_date field

Zlaticanin:
In Vault 1.18, we added the `explicit_max_ttl` field and began storing the `key_end_date` field in the internal data to support it. This `key_end_date` is required for renewal logic to correctly calculate how long a service principal key remains valid. 

However, leases created in Vault <= 1.18 don't include this field, which caused the renewals on those older leases to fail with 
```
Error making API request.  URL: PUT http://127.0.0.1:8200/v1/sys/leases/renew Code: 400. Errors:  * failed to renew entry: resp: (*logical.Response)(nil) err: internal data 'key_end_date' not found 
```
This PR addresses the issue by introducing a fallback behavior where we check if a lease is missing a `key_end_date` field. If it does, we assume it was created before 1.18 and we calculate a end date using `IssueTime` + `spExpiration` (system default of 10 years).

### DIFF
--- a/path_service_principal.go
+++ b/path_service_principal.go
@@ -237,11 +237,22 @@ func (b *azureSecretBackend) spRenew(ctx context.Context, req *logical.Request, 
 		return nil, nil
 	}
 
-	// Determine remaining lifetime of SP secret in Azure
-	keyEndDateRaw, ok := req.Secret.InternalData["key_end_date"]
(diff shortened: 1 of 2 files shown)


fairclothjm (approved):
LGTM

